PR #1, opened 2017-06-13
Make a briefs frontend

allait:
### Replace buyer-frontend mentions with briefs-frontend

Replaces the docker container name, repository references and config
variable values.

### Rename buyers blueprint to main

To keep it closer to the other apps, the majority of the routes
are under the `main` blueprint.

This also adds a copy of `main/errors.py` from the buyer frontend
with the common application error handlers.

### Fix test login helper

Now that login view is not part of the app we need to use the same
approach supplier-frontend is using for logging in users during a
test run.

### Add external blueprint for linking to pages from other frontend apps

Templates are using url_for with flask view names to generate links
to other application pages. After the split, some of these pages no
longer exist in the same application.

Normally, this would mean that we need to replace the `url_for` calls
with inline URLs. Instead, we add a new `external` blueprint without
the app's url prefix.

This allows us to copy the view definitions from the buyer-frontend
and keep using `url_for` with the new `external.*` view names.

The views aren't actually called, since the link will be routed to
a different app, but we get all the safety benefits of using `url_for`
instead of hardcoded URLs.

In the future, we could move the external view and blueprint to utils
so it can be shared between the apps.


app/templates/_footer_categories.html
```diff
@@ -24,7 +24,7 @@
         <a href="https://www.gov.uk/guidance/the-g-cloud-framework-on-the-digital-marketplace">G-Cloud framework</a>
       </li>
       <li>
-        <a href="{{ url_for('main.suppliers_list_by_prefix') }}">G-Cloud supplier A–Z</a>
```
pcraig3: I guess we're not using an external blueprint for this because there aren't any query parameters?

app/templates/_footer_categories.html
```diff
@@ -24,7 +24,7 @@
...
+        <a href="/g-cloud/suppliers">G-Cloud supplier A–Z</a>
```
pcraig3: I guess we're not using an external blueprint for this because there aren't any query parameters?

allait: We've taken this from the supplier frontend. We could use external blueprint approach for this, but it feels like there's an argument for sharing this and the header file across apps to begin with.

tests/helpers.py
```diff
@@ -12,6 +13,27 @@
...
+
+
+login_for_tests = Blueprint('login_for_tests', __name__)
+
+
+@login_for_tests.route('/auto-buyer-login')
+def auto_buyer_login():
+    user_json = {"users": {
+        'id': 123,
+        'name': u'Ā Buyer',
```
idavidmcdonald: oooo unicode

idavidmcdonald:
Oh and good work on keeping the git history! How did you do that?

Makefile
```diff
@@ -50,11 +50,11 @@ show_environment:
 docker-build:
 	$(if ${RELEASE_NAME},,$(eval export RELEASE_NAME=$(shell git describe)))
 	@echo "Building a docker image for ${RELEASE_NAME}..."
-	docker build --pull -t digitalmarketplace/buyer-frontend --build-arg release_name=${RELEASE_NAME} .
-	docker tag digitalmarketplace/buyer-frontend digitalmarketplace/buyer-frontend:${RELEASE_NAME}
+	docker build -t digitalmarketplace/briefs-frontend --build-arg release_name=${RELEASE_NAME} .
```
allait: It doesn't add much now that we're versioning the base image

allait:
@idavidmcdonald we didn't do anything smart with commit history, it's a clone of the buyer frontend repo, so it also contains commits related to the removed files.

requirements.txt
```diff
@@ -9,9 +9,3 @@ odfpy==1.3.3
 git+https://github.com/alphagov/digitalmarketplace-utils.git@26.2.1#egg=digitalmarketplace-utils==26.2.1
 git+https://github.com/alphagov/digitalmarketplace-content-loader.git@4.4.1#egg=digitalmarketplace-content-loader==4.4.1
 git+https://github.com/alphagov/digitalmarketplace-apiclient.git@8.8.0#egg=digitalmarketplace-apiclient==8.8.0
-
```
allait: Yeah, we can drop it everywhere. We don't need these dependency for Docker deployments.

PR #3, opened 2017-06-16
Add the `create_buyer_account` views from the buyer app.

Wynndow:
### Add `create` Blueprint and views  …
The buyer app had a couple of views, `create_buyer_account` and
`submit_create_buyer_account` both with the url `/buyers/create`.

This URL meant they clashed with the new briefs app and so when trying
to access them a user would get a 404.

These views don't actually create a User themselves, they just send an
email with a token. Actual user creation is still dealt with on the
buyer app. As the only need to create a buyer user is to create a
brief, it feels right that if these views need to move they should go
with the briefs app.

A new blueprint, `create` has been created. This is because the `buyers`
blueprint is behind `login_required`, and putting these routes behind
the `dos` blueprint feels wrong.

### Add external links to buyer app for user creation  …
The new views for creating buyer accounts need to link back to the buyer
app for actually creating users. The necessary links are added here in
the same way as other external links


app/main/__init__.py
```diff
@@ -4,6 +4,7 @@
 
 main = Blueprint('buyers', __name__)
 dos = Blueprint('dos', __name__)
+create = Blueprint('create', __name__)
```
allait: `create` feels a bit too general. Maybe `create_user` or `create_buyer`?

app/main/helpers/hash_email.py
```diff
@@ -0,0 +1,9 @@
+import hashlib
+import base64
+
+
+def hash_email(email):
```
allait: I think we have a copy of this function in utils (https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/email/helpers.py). Can we use that?

allait:
Oh and if there's a separate blueprint for these endpoints then maybe we should put the view and the test files in related directories - `views/create_.../...` and `tests/create.../...` instead of `*/main/...`

PR #4, opened 2017-06-20
Bring in frontend toolkit updates.

carolinegreen:
Bring in new no javascript buttons.  Also add GOV.UK elements as a dependency and alter gulp build process to allow this.

Relies on: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/353


idavidmcdonald:
Am closing this for the moment. There are potentially some issues that were picked up in a local QA and this needs a proper look into before we know it's good to go.

The briefs team has a ticket in our backlog to look at this properly and upgrade.

PR #8, opened 2017-07-14
Pin requirements

allait:
requirements.txt is generated using a Python 3 virtualenv to keep the py2 backports of Python 3 stdlib packages (like `enum34` and `ipaddress`) out. This leaves those packages unpinned, but shouldn't cause any issues otherwise.

### Reorganize Makefile and requirements files

Renames requirements.txt to requirements-app.txt and requirements_for_test.txt to requirements-dev.txt.

Removes any unused dependencies or dependencies provided by other packages since those will be listed in the full requirements.txt.

Updates Makefile to add freeze-requirements and rename targets to be more consistent with other apps.

(see alphagov/digitalmarketplace-api#615 for details)


Makefile
```diff
@@ -8,53 +8,83 @@ else
...
 	npm install
 
-frontend_build:
+.PHONY: frontend-build
+frontend-build:
 	npm run --silent frontend-build:${GULP_ENVIRONMENT}
 
-test: show_environment frontend_build test_pep8 test_python test_javascript
+.PHONY: test
+test: show-environment test-requirements frontend-build test-pep8 test-python test-javascript
```
Wynndow: Is it worth adding `requirements-dev` to this?

Makefile
```diff
@@ -8,53 +8,83 @@ else
...
+.PHONY: requirements
+requirements: virtualenv test-requirements upgrade-pip requirements.txt
 	${VIRTUALENV_ROOT}/bin/pip install -r requirements.txt
 
-requirements_for_test: virtualenv upgrade_pip requirements_for_test.txt
-	${VIRTUALENV_ROOT}/bin/pip install -r requirements_for_test.txt
-
-npm_install: package.json
+.PHONY: requirements-dev
+requirements-dev: virtualenv test-requirements requirements-dev.txt
```
Wynndow: Would `requirements-test` be a better name for this (and the requirements file)? Bit of a pain to change them all now though...

Makefile
```diff
@@ -8,53 +8,83 @@ else
...
-test_pep8: virtualenv
+.PHONY: test-pep8
+test-pep8: virtualenv
 	${VIRTUALENV_ROOT}/bin/pep8 .
 
-test_python: virtualenv
-	${VIRTUALENV_ROOT}/bin/py.test ${PYTEST_ARGS}
+.PHONY: test-python
+test-python: virtualenv
+	${VIRTUALENV_ROOT}/bin/py.test ${PYtest-ARGS}
```
Wynndow: Lower case used here for the variable, but all upper in the `.travis.yml`

Makefile
allait: That's what it used to be (`requirements_for_test.txt`). The reason for the rename was to reflect the fact that this should also be used for non-test related dev dependencies - like script dependencies or other tools we use that shouldn't be installed in live environments.

Makefile
allait: We kept them as separate steps so far. We could make an argument about `make test` running pip install each time now that the versions are pinned, but it feels like a separate conversation to me.

Makefile
allait: Oops! should be fixed now

PR #15, opened 2017-08-24
Add create-your-account-complete endpoint and template

allait:
Most of the buyer account creation flow is part of the briefs frontend
at the moment, but since the "check your email" page URL wasn't prefixed
with `/buyers` it was left in the buyer frontend.

This moves it to the rest of the buyer account creation pages and
changes the URL to be prefixed with `/buyers/`.


TheDoubleK:
Is this not also the same page that suppliers end up at after creating their account?

allait:
@TheDoubleK supplier frontend has a separate page with it's own template at `/suppliers/create-your-account-complete`.

PR #20, opened 2017-08-31
Buyer account page

DilwoarH:
As a G-Cloud buyer
I need to be able to return to a list of the projects i am running
So that I can keep track of my projects

Ticket: https://trello.com/c/FiR3NLyZ/679-buyer-account-page

 
buyer-frontend


DilwoarH:
Feature flag now working

tests/main/views/test_buyers.py
```diff
@@ -92,7 +92,7 @@ def setup_method(self, method):
     def test_draft_briefs_section(self, data_api_client, find_briefs_mock):
         data_api_client.find_briefs.return_value = find_briefs_mock
 
-        res = self.client.get("/buyers")
+        res = self.client.get("/buyers/dos-requirements")
         tables = html.fromstring(res.get_data(as_text=True)).xpath('//table')
```
DilwoarH: @samuelhwilliams / @galund - could I just do the following

```
buyer_dashboard_url = "/buyers"

import flask_featureflags;
if flask_featureflags.is_active('DIRECT_AWARD_PROJECTS'):
    buyer_dashboard_url = "/buyers/dos-requirements"
```

samuelhwilliams: Doing it that way is fine by me.

app/assets/scss/_index-page.scss
```diff
@@ -155,3 +155,7 @@
     }
   }
 }
+
+.heading-xmedium {
```
galund: Our classes should be named for something vaguely semantic - not to do with how they look.

And in fact, our H3's (and H2's, etc) should be a consistent size across the site. So this shouldn't really be a thing... should it?

app/assets/scss/_index-page.scss
```diff
@@ -155,3 +155,7 @@
...
+  @extend %heading-xmedium;
+}
```
galund: newline :)

app/templates/buyers/index.html
```diff
@@ -0,0 +1,56 @@
...
+    }
+    ]
+  %}
+    {% include "toolkit/breadcrumb.html" %}
+  {% endwith %}
+{% endblock %}
+
+{% block main_content %}
+
+    <div class="grid-row">
```
galund: 2 spaces for HTML pls

tests/main/views/test_buyers.py
```diff
@@ -106,7 +106,7 @@ def setup_method(self, method):
     def test_draft_briefs_section(self, data_api_client, find_briefs_mock):
         data_api_client.find_briefs.return_value = find_briefs_mock
 
-        res = self.client.get("/buyers")
+        res = self.client.get("/buyers/requirements/digital-outcomes-and-specialists")
```
galund: I wonder if we could use url_for here, it might not fit in with existing tests but it'd save us some grief down the line

app/assets/scss/_index-page.scss
TheDoubleK: Yep we should avoid random per-app styling and stick to the toolkit-defined styles where possible.

Also, yes to semantic naming.  Not even just "vaguely" if possible.

TheDoubleK: Probably "Smaller" heading from here? http://alphagov.github.io/digitalmarketplace-frontend-toolkit/page-heading.html

galund: "smaller" is definitely not semantic _either_  (so I think my "vaguely" wasn't so far off ;-) )

TheDoubleK: "Heading" is the semantics, "Smaller" is an attribute of that thing :P

galund: hmm, the problem here though is not "heading", it's "medium", which is as much a (formatting-related) attribute as is "small"... but more to the point, what is a "small h2" compared to a "regular h3"? How could that help users understand the structure of our pages...

tests/helpers.py
```diff
@@ -66,6 +66,10 @@ def setup_method(self, method):
         self.client = self.app.test_client()
         self.get_user_patch = None
 
+        self.buyer_dashboard_url = "/buyers"
```
galund: Thanks.. this variable should be named buyer_requirements_dashboard_url I think based on what it's actually containing?

app/assets/scss/_index-page.scss
TheDoubleK: Yes, point taken :)
I think the main thing is that we should be using frontend toolkit elements for everything we possibly can.

DilwoarH: I can call it `subheading` ?

tests/helpers.py
DilwoarH: sure I'll get that updated, `briefs_dashboard_url` sounds more consistent.

app/templates/buyers/index.html
```diff
@@ -0,0 +1,60 @@
...
+        heading = current_user.name
+    %}
+    {% include 'toolkit/page-heading.html' %}
+    {% endwith %}
+  </div>
+  <div class="column-full dm-account-overview-options">
+    
+    <div class="framework">
+      <div class="dmspeak">
+        <h3 class="heading-xmedium">Cloud hosting, software and support</h3>
```
galund: Let's lose the class completely if we don't need it (also a few lines down).

DilwoarH: unfortunately, we can't. Ref: http://alphagov.github.io/digitalmarketplace-frontend-toolkit/shared_scss/dmspeak.html

galund: oh, lol, you could have just told @TheDoubleK and me that this was an existing class name then :)

(But still gahhh)

app/templates/buyers/index.html
```diff
@@ -0,0 +1,60 @@
...
+      </div>
+      <p>Buy cloud technology through the G-Cloud framework.</p>
```
galund: This content is weird, but it's certainly fine for us to test with

app/templates/buyers/index.html
TheDoubleK: Ha! But still, if it's a style in the toolkit it should be imported from there, not copied into the app.

app/templates/buyers/index.html
samuelhwilliams: Cloud hosting software and support should be pulled from frameworks

TheDoubleK: Yep, ideally no frameworks should be explicitly named in our front-end apps.

TheDoubleK: Apps need to be applicable to any future similar frameworks.

config.py
```diff
@@ -102,6 +103,7 @@ class Test(Config):
     SECRET_KEY = "KEY"
 
     FEATURE_FLAGS_NEW_SUPPLIER_FLOW = enabled_since('2016-11-29')
+    FEATURE_FLAGS_DIRECT_AWARD_PROJECTS = True
```
samuelhwilliams: You should keep this as the `enabled_since` thing we use elsewhere to remain consistent. It should still pass the boolean check you do later.

tests/main/views/test_buyers.py
```diff
@@ -241,23 +239,27 @@ def test_flash_message_shown_if_brief_has_just_been_updated(self, data_api_clien
...
-            assert res.location == 'http://localhost/user/login?next=%2Fbuyers'
+            assert res.location == 'http://localhost/user/login?next={}'.format(
+                self.briefs_dashboard_url.replace('/', '%2F')
+            )
 
     def test_supplier_cannot_access_buyer_pages(self):
         with self.app.app_context():
             self.login_as_supplier()
-            res = self.client.get('/buyers')
+            res = self.client.get(self.briefs_dashboard_url)
```
samuelhwilliams: This test should probably cover both pages: the main (new) buyers dashboard, and the dos buyers dashboard.

app/templates/buyers/index.html
TheDoubleK: `This content is weird` :+1:
For years we have tried to remove the word "framework" as much as possible from the marketplace, the rationale being that buyers come to buy a thing and shouldn't need to know or understand the concept of frameworks.  Instead we should help them find the thing they want to buy.

Unless there has been a complete U-turn in policy around this for the marketplace in general then I'd be wary of even putting this in front of people in labs.

app/templates/buyers/index.html
```diff
@@ -0,0 +1,60 @@
...
+          You don't have any saved searches
+        {% endif %}
+      </p>
+    </div>
+
+    <div class="framework">
+      <div class="dmspeak">
+        <h3 class="heading-xmedium">Digital outcomes, specialists and user research</h3>
+      </div>
+      <p>Buy services through the Digital Outcomes and Specialists framework.</p>
```
TheDoubleK: Same here - the briefs pages don't mention "framework" anywhere, we really shouldn't be introducing it into content at all.

config.py
TheDoubleK: Any "truthy" value will turn the feature flag on - so *anything* other than `False` or `None` I think will work.

We do the `enabled_since` thing so we can see on the status pages for different environments which features are on and since when (e.g. see https://www.preview.marketplace.team/_status etc.)

TheDoubleK:
```
"flags": {
"FEATURE_FLAGS_TRANSACTION_ISOLATION": "2015-08-27"
}
```

app/templates/buyers/index.html
DilwoarH: it has been imported now - no longer using copier @TheDoubleK

config.py
DilwoarH: @samuelhwilliams @TheDoubleK - enabled_since always passed in False when running tests. Never evaluated the actual date.

app/templates/buyers/index.html
```diff
@@ -33,7 +33,7 @@ <h3 class="heading-xmedium">Cloud hosting, software and support</h3>
       <p>Buy cloud technology through the G-Cloud framework.</p>
       <p>
         {% if user_projects_total %}
-          <a href="/buyers/direct-award">View your saved searches</a><br>
+          <a href="/buyers/direct-award/g-cloud">View your saved searches</a><br>
```
samuelhwilliams: We shouldn't be hard-coding g-cloud here!

app/templates/buyers/index.html
```diff
@@ -0,0 +1,60 @@
...
+  <div class="column-full dm-account-overview-options">
+    
+    <div class="framework">
+      <div class="dmspeak">
+        <h3 class="heading-xmedium">Cloud hosting, software and support</h3>
+      </div>
+      <p>Buy cloud technology through the G-Cloud framework.</p>
+      <p>
+        {% if user_projects_total %}
+          <a href="/buyers/direct-award/g-cloud">View your saved searches</a><br>
```
samuelhwilliams: We shouldn't be hardcoding `g-cloud` here again.

app/templates/buyers/index.html
galund: We don't have @karlchillmaid available right now, so this will have to go to labs as it is. We'll keep the ticket open until the content is sorted.

app/templates/buyers/index.html
```diff
@@ -0,0 +1,60 @@
...
+    </div>
+
+    <div class="framework">
+      <div class="dmspeak">
+        <h3 class="heading-xmedium">Digital outcomes, specialists and user research</h3>
+      </div>
+      <p>Buy services through the Digital Outcomes and Specialists framework.</p>
+      <p>
+        {% if user_briefs_total %}
+          <a href="/buyers/requirements/digital-outcomes-and-specialists">View your requirements</a><br>
```
idavidmcdonald: Should we be using `url_for` here for these links?

app/main/__init__.py
```diff
@@ -19,6 +20,7 @@
 
 
 @main.before_request
+@flask_featureflags.is_active_feature('DIRECT_AWARD_PROJECTS')
```
idavidmcdonald: Given when this gets deployed, the feature flag will not be active for STAGING and PRODUCTION, does this mean that we are turning off `require_login` for those environments (or maybe 404ing)? Or have I misunderstood how the feature flags interact with this function?

https://github.com/trustrachel/Flask-FeatureFlags#controllersviews

galund: I'd certainly hope that they are entirely independent checks, i.e. this view is only accessible if the feature is enabled AND the user is logged in. If there's any doubt about that we'd better check!

app/templates/buyers/index.html
samuelhwilliams: `Cloud hosting, software and support` is framework-specific content and should probably come via the content loader, but you may want to do this after labs.

app/main/views/buyers.py
```diff
@@ -35,6 +37,24 @@
 
 @main.route('')
 def buyer_dashboard():
+
```
samuelhwilliams: This extra newline after the `def` weird, please remove it.

app/main/views/buyers.py
```diff
@@ -35,6 +37,24 @@
...
+    if flask_featureflags.is_active('DIRECT_AWARD_PROJECTS'):
+
```
samuelhwilliams: Same but (albeit marginally less weird)

app/main/__init__.py
```diff
@@ -1,6 +1,7 @@
 from flask import Blueprint, current_app, flash
 from flask_login import current_user, login_required
 from dmcontent.content_loader import ContentLoader
+import flask_featureflags
```
galund: This is probably not needed now?

tests/helpers.py
```diff
@@ -66,6 +66,10 @@ def setup_method(self, method):
         self.client = self.app.test_client()
         self.get_user_patch = None
 
+        self.briefs_dashboard_url = "/buyers"
+        if self.app.config.get('FEATURE_FLAGS_DIRECT_AWARD_PROJECTS'):
+            self.briefs_dashboard_url = "/buyers/requirements/digital-outcomes-and-specialists"
+
     def teardown_method(self, method):
```
benvand: This doesn't live on the main tests base class if it's only ever used in `test_buyers.py`. We want to keep base classes as slim as possible and attach attributes like these to classes specific to the tests that use them.

This would also be more pythonic as a ternary assignment.
http://book.pythontips.com/en/latest/ternary_operators.html

app/main/views/buyers.py
```diff
@@ -35,6 +37,22 @@
...
+        user_projects_total = len(data_api_client.find_direct_award_projects(current_user.id).get('projects', []))
+
+        return render_template(
+            'buyers/index.html',
+            user_briefs_total=user_briefs_total,
+            user_projects_total=user_projects_total
+        )
+
+    else:
+        return buyer_dos_requirements()
```
benvand: Why do we want 2 end points showing exactly the same content? This is not a pattern we follow elsewhere.
I'd suggest a redirect and use the original url unless there's a really good reason not to.

```
return redirect(
    url_for(".buyer_dos_requirements",)
)
```

galund: We considered it didn't make sense to change the URLs, not to change the breadcrumbs. So when the new feature goes live, the URLs change and the breadcrumbs change at the same time (and no redirect is required, because both URLs still exist).

tests/helpers.py
DilwoarH: il get this moved.

DilwoarH: Oh wait, there was a reason why it had to be there. For some reason the self.apps couldn't see the config attribute inside of the subclasses. I will look to reimplement it and see if it works.

app/main/views/buyers.py
benvand: > So when the new feature goes live, the URLs change

Do you mean the views?
Confused by this 😕

galund: Well they kinda both change: `/buyers/` effectively switches from being a DOS-specific view to being a general one, and the existing DOS-specific view moves to the new URL. We don't want any of this to show up until the feature flag is turned on.

benvand: That means we're going to need to check every link to `'/buyers'` and `url_for('buyer_dashboard')` in the app to make sure they still make sense in context right?

galund: Right :) https://trello.com/c/mF0NJvQD/717-find-links-to-buyer-account-page-check-dos-impact

galund: See also https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/30

benvand: Cool, so until the feature flag is on nothing changes. 
(`'/buyers'`, `buyer_dashboard `) point to where it always have
(`'/requirements/digital-outcomes-and-specialists'`, `buyer_dos_requirements `) points to the buyer dash too.

When the feature flag is on the new dash is up.

benvand: This would have been better achieved by registering the urls based on the feature flag. Then we would have had no duplicate page and could have avoided the if statement that skips a whole view:
```

def buyer_dashboard():
    user_briefs_total = len(data_api_client.find_briefs(current_user.id).get('briefs', []))
    user_projects_total = len(data_api_client.find_direct_award_projects(current_user.id).get('projects', []))

    return render_template(
        'buyers/index.html',
        user_briefs_total=user_briefs_total,
        user_projects_total=user_projects_total
    )


def buyer_dos_requirements():
    user_briefs = data_api_client.find_briefs(current_user.id).get('briefs', [])

    draft_briefs = sorted(
        add_unanswered_counts_to_briefs([brief for brief in user_briefs if brief['status'] == 'draft'], content_loader),
        key=lambda i: datetime.strptime(i['createdAt'], DATETIME_FORMAT),
        reverse=True
    )
    live_briefs = sorted(
        [brief for brief in user_briefs if brief['status'] == 'live'],
        key=lambda i: datetime.strptime(i['publishedAt'], DATETIME_FORMAT),
        reverse=True
    )
    closed_briefs = sorted(
        [brief for brief in user_briefs if brief['status'] in CLOSED_BRIEF_STATUSES],
        key=lambda i: datetime.strptime(i['applicationsClosedAt'], DATETIME_FORMAT),
        reverse=True
    )

    breadcrumbs = [
        {
            "link": "/",
            "label": "Digital Marketplace"
        }
    ]

    if flask_featureflags.is_active('DIRECT_AWARD_PROJECTS'):
        breadcrumbs += [
            {
                "link": url_for("buyers.buyer_dashboard"),
                "label": "Your account"
            }
        ]

    return render_template(
        'buyers/dashboard.html',
        draft_briefs=draft_briefs,
        live_briefs=live_briefs,
        closed_briefs=closed_briefs,
        breadcrumbs=breadcrumbs,
    )


# Set routes depending on whether direct awards feature is active.
if flask_featureflags.is_active('DIRECT_AWARD_PROJECTS'):
    main.route('', endpoint='buyer_dashboard')(buyer_dos_requirements)
    main.route(
        '/requirements/digital-outcomes-and-specialists',
        endpoint='buyer_dos_requirements'
    )(buyer_dos_requirements)
else:
    main.route('', endpoint='buyer_dashboard')(buyer_dos_requirements)
    main.route('', endpoint='buyer_dos_requirements')(buyer_dos_requirements)

```

PR #22, opened 2017-09-04
845-bug-buyer-requirements-are-ordered-wrong

benvand:
Add ordering to buyer dashboard
https://trello.com/c/Qs4pUyqO/845-bug-buyer-requirements-are-ordered-oldest-to-newest-in-both-tables-on-their-dashboard-this-is-wrong-they-should-be-ordered-newes

### DIFF
--- a/app/main/views/buyers.py
+++ b/app/main/views/buyers.py
@@ -22,6 +22,7 @@ from ..forms.awards import AwardedBriefResponseForm
 
 from dmapiclient import HTTPError
 from dmutils.dates import get_publishing_dates
+from dmutils.formats import DATETIME_FORMAT
 from dmutils import csv_generator
 from datetime import datetime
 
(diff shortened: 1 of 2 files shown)


app/templates/buyers/dashboard.html
```diff
@@ -54,7 +54,7 @@
 
 {{ summary.heading("Unpublished requirements", id="unpublished_requirements") }}
 {% call(item) summary.list_table(
-    draft_briefs,
+    draft_briefs|sort(attribute='createdAt', reverse=True),
```
katstevens: I would have a slight preference for doing the sort in the backend rather than the template. I doubt we'd gain much speed-wise but as we're splitting out the briefs into separate lists in the backend, does it make sense to keep the ordering logic there as well?

benvand:
@CrystalPea 
Yeah, would have used `withdrawnAt` but it's not included in the serialization.

benvand:
@katstevens Done

PR #25, opened 2017-09-07
Redirect to DOS Dashboard

samuelhwilliams:
## Summary
We've moved the requirements dashboard to another URL but didn't update where the 'award a brief' and 'delete a requirement' flows end. These need to be updated to point to the correct dashboard because the briefs-frontend uses old-style flash messages, and the buyer-frontend uses new-style (frontend-toolkit) flash messages. The buyer-frontend doesn't know how to render the flash messages generated by the briefs-frontend and the functional tests fail.

Also it feels like awarding a brief/deleting a requirement intend to go back to the DOS dashboard, so this feels like the right change regardless.

However, if we don't want to make these go straight to the DOS dashboard, we could simply have extra steps in the functional tests to click on 'View your requirements' from the dual buyers dashboard, and the flash messages should then be picked up and displayed. Open to input on which we prefer.

### DIFF
--- a/app/main/views/buyers.py
+++ b/app/main/views/buyers.py
@@ -484,6 +484,10 @@ def award_brief_details(framework_slug, lot_slug, brief_id, brief_response_id):
             ), 400
 
         flash({"updated-brief": brief.get("title")})
+
+        if flask_featureflags.is_active('DIRECT_AWARD_PROJECTS'):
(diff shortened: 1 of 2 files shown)


samuelhwilliams:
Let's address the breadcrumbs separately as we should get input from content.

PR #26, opened 2017-09-07
816 3 collect status when buyer has cancelled procurement no journey frontend

benvand:
https://trello.com/c/Umt24W6d/816-3-collect-status-when-buyer-has-cancelled-procurement-no-journey-frontend

### DIFF
--- a/app/main/forms/cancel.py
+++ b/app/main/forms/cancel.py
@@ -1,0 +1,22 @@
+from flask.ext.wtf import Form
+from wtforms import RadioField, validators
+
+
+class CancelBriefForm(Form):
+    """Form for the buyer to tell us why they cancel the requirement
+    """
+    choices = [
+        ('cancel', 'The requirement has been cancelled'),
+        ('unsuccessful', 'There were no suitable suppliers'),
+    ]
+    cancel_reason = RadioField(
+        validators=[validators.InputRequired(message='You need to answer this question.')],
+        choices=choices
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(CancelBriefForm, self).__init__(*args, **kwargs)
+        brief = kwargs.pop('brief')
+        brief_name = brief.get('title', brief.get('lotName', ''))
+        self.cancel_reason.label.text = "Why do you need to cancel {}?".format(brief_name)
+        self.cancel_reason.toolkit_macro_options = [{'value': i[0], 'label': i[1]} for i in self.choices]
--- a/app/main/views/buyers.py
+++ b/app/main/views/buyers.py
@@ -19,6 +19,7 @@ from ..helpers.buyers_helpers import (
 
 from ..helpers.ods import SpreadSheet
 from ..forms.awards import AwardedBriefResponseForm
+from ..forms.cancel import CancelBriefForm
 
 from dmapiclient import HTTPError
 from dmutils.dates import get_publishing_dates
@@ -452,6 +453,60 @@ def award_brief(framework_slug, lot_slug, brief_id):
         brief=brief,
         form=form
     ), 200
+
+
+@main.route('/frameworks/<framework_slug>/requirements/<lot_slug>/<brief_id>/cancel', methods=['GET', 'POST'])
+def cancel_brief(framework_slug, lot_slug, brief_id):
+    form = None
+    errors = {}
+    get_framework_and_lot(
+        framework_slug,
+        lot_slug,
+        data_api_client,
+        allowed_statuses=['live', 'expired'],
+        must_allow_brief=True,
+    )
+    brief = data_api_client.get_brief(brief_id)["briefs"]
+    if not is_brief_correct(brief, framework_slug, lot_slug, current_user.id):
+        abort(404)
+    if brief["status"] != "closed":
+        abort(404)
+
+    if request.method == "POST":
+        form = CancelBriefForm(request.form, brief=brief)
+        if not form.validate_on_submit():
+            errors = {
+                key: {'question': form[key].label.text, 'input_name': key, 'message': form[key].errors[0]}
+                for key, value in form.errors.items()
+            }
+        else:
+            new_status = form.data.get('cancel_reason')
+            try:
+                if new_status == 'cancel':
+                    data_api_client.cancel_brief(
+                        brief_id,
+                        user=current_user.email_address
+                    )
+                elif new_status == 'unsuccessful':
+                    data_api_client.update_brief_as_unsuccessful(
+                        brief_id,
+                        user=current_user.email_address
+                    )
+                else:
+                    abort(400, "Unrecognized status '{}'".format(new_status))
+                flash({"updated-brief": brief.get("title")})
+                return redirect(
+                    url_for('.view_brief_overview', framework_slug=framework_slug, lot_slug=lot_slug, brief_id=brief_id)
+                )
+            except HTTPError as e:
+                abort(500, "Unexpected API error when cancelling brief")
+
+    return render_template(
+        "buyers/cancel_brief.html",
+        brief=brief,
+        form=form or CancelBriefForm(brief=brief),
+        errors=errors,
+    ), 200 if not errors else 400
 
 
 @main.route(
--- a/app/templates/buyers/brief_overview.html
+++ b/app/templates/buyers/brief_overview.html
@@ -53,6 +53,28 @@
         {% endif %}
       {% endwith %}
     {% endblock %}
+
+    {% with messages = get_flashed_messages(with_categories=true) %}
+        {% if messages %}
+            {% for category, message in messages %}
+                {% if category == 'error' %}
+                    <div class="banner-destructive-without-action">
+                {% else %}
+                    <div class="banner-success-without-action">
+                {% endif %}
+                        <p class="banner-message">
+                {% if 'updated-brief' in message %}
+                        You've updated '{{ message.get('updated-brief') }}'
+                {% else %}
+                    {{ message }}
+                {% endif %}
+                        </p>
+                    </div>
+            {% endfor %}
+        {% endif %}
+    {% endwith %}
+
+>>>>>>> a5fee71... Update brief summary with link to cancel and banner on update
   </div>
 
   <div class="grid-row">
@@ -271,12 +293,17 @@
     </div>
     <div class="column-one-third">
 <<<<<<< HEAD
+<<<<<<< HEAD
       {% if brief.publishedAt %}
         {% include 'buyers/_requirements_meta.html' %}
 =======
       {% if brief.status == 'closed' %}
         <a href="{{ url_for('buyers.cancel_brief', framework_slug=framework.slug, lot_slug=brief.lotSlug, brief_id=brief.id) }}">Cancel requirement</a>
 >>>>>>> 0a5a78a... temp
+=======
+      {% if brief.status == 'closed' and framework.status in ['live', 'expired'] %}
+        <a href="{{ url_for('buyers.cancel_brief', framework_slug=framework.slug, lot_slug=brief.lotSlug, brief_id=brief.id) }}">Cancel requirement</a>
+>>>>>>> a5fee71... Update brief summary with link to cancel and banner on update
       {% endif %}
     </div>
   </div>
--- a/app/templates/buyers/brief_overview.html
+++ b/app/templates/buyers/brief_overview.html
@@ -53,28 +53,6 @@
         {% endif %}
       {% endwith %}
     {% endblock %}
-
-    {% with messages = get_flashed_messages(with_categories=true) %}
-        {% if messages %}
-            {% for category, message in messages %}
-                {% if category == 'error' %}
-                    <div class="banner-destructive-without-action">
-                {% else %}
-                    <div class="banner-success-without-action">
-                {% endif %}
-                        <p class="banner-message">
-                {% if 'updated-brief' in message %}
-                        You've updated '{{ message.get('updated-brief') }}'
-                {% else %}
-                    {{ message }}
-                {% endif %}
-                        </p>
-                    </div>
-            {% endfor %}
-        {% endif %}
-    {% endwith %}
-
->>>>>>> a5fee71... Update brief summary with link to cancel and banner on update
   </div>
 
   <div class="grid-row">
@@ -100,7 +78,7 @@
                 'awarded': 'Done',
                 'cancelled': 'Done',
                 'unsuccessful': 'Done',
-              },git st
+              },
             },
             {
               'title': 'Set how you’ll evaluate suppliers',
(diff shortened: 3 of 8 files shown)


samuelhwilliams:
Will need to run and commit `make freeze-requirements` as a first step.

benvand:
@samuelhwilliams Can you let me know if I've done that right?

samuelhwilliams:
Yep that seems fine to me.

app/templates/buyers/brief_overview.html
```diff
@@ -38,7 +38,28 @@
           </form>
         </div>
       {% endif %}
-   {% endblock %}
+    {% endblock %}
+
+    {% with messages = get_flashed_messages(with_categories=true) %}
```
katstevens: Could the flash banner go in the `{% block before_heading %}` above, alongside the `delete_requested` banner? (Genuine question - I'm not sure how strict we are about what goes in content blocks)

app/main/forms/cancel.py
```diff
@@ -0,0 +1,22 @@
+from flask.ext.wtf import Form
+from wtforms import RadioField, validators
+
+
+class CancelBriefForm(Form):
```
katstevens: Curious - what was behind the decision to use WTF instead of content loader in the end? Was it validation/errors?

app/templates/buyers/brief_overview.html
katstevens: We've also got an almost identical bit of code in the `dashboard.html` template. Any way we could make this more DRY for future re-use? ('No' is an acceptable answer to this question!)

tests/main/views/test_buyers.py
```diff
@@ -1871,6 +1871,56 @@ def test_show_awarded_brief_summary_page_for_live_and_expired_framework(self, da
...
+
+        res = self.client.get(
+            "/buyers/frameworks/digital-outcomes-and-specialists/requirements/digital-specialists/1234"
+        )
+
+        page_html = res.get_data(as_text=True)
+        document = html.fromstring(page_html)
+
+        assert res.status_code == 200
+        assert "Cancel requirement" in document.text_content()
```
katstevens: Rather than having these separate two tests, could we add in 1-line assertions to the first 5 tests in `TestBriefSummaryPage`?

tests/main/views/test_buyers.py
```diff
@@ -3676,3 +3642,122 @@ def test_award_brief_details_post_raises_400_and_displays_error_messages_and_pre
...
+    def test_redirect_on_successful_status_change(self, status):
+        resp = self.client.post(
+            self.url.format(brief_id=123), data={'cancel_reason': status}
+        )
+
+        redirect_text = html.fromstring(resp.get_data(as_text=True)).text_content().strip()
+        expected_url = '/buyers/frameworks/digital-outcomes-and-specialists-2/requirements/digital-outcomes/123'
+
+        assert resp.status_code == 302
+        assert expected_url in redirect_text
```
katstevens: Can we assert here whether the flash message is displayed (as we do in the awards test class - `self.assert_flashes("updated-brief")`)?

app/main/forms/cancel.py
```diff
@@ -0,0 +1,22 @@
...
+    """Form for the buyer to tell us why they cancel the requirement
+    """
+    choices = [
+        ('cancel', 'The requirement has been cancelled'),
+        ('unsuccessful', 'There were no suitable suppliers'),
+    ]
+    cancel_reason = RadioField(
+        validators=[validators.InputRequired(message='Answer required')],
```
idavidmcdonald: Might be worth double checking with content if this is our default message (if you haven't already). I swear I've seen 'You must answer this question' elsewhere

app/main/views/buyers.py
```diff
@@ -444,6 +445,60 @@ def award_brief(framework_slug, lot_slug, brief_id):
...
+                elif new_status == 'unsuccessful':
+                    data_api_client.update_brief_as_unsuccessful(
+                        brief_id,
+                        user=current_user.email_address
+                    )
+                else:
+                    abort(400, "Unrecognized status '{}'".format(new_status))
+                flash({"updated-brief": brief.get("title")})
+                return redirect(
+                    url_for('.view_brief_overview', framework_slug=framework_slug, lot_slug=lot_slug, brief_id=brief_id)
```
idavidmcdonald: It feels weird that if you award a brief you go to the list of all your briefs. If you cancel a brief then you go to your brief overview. Might be worth checking with Laurence about this

app/templates/buyers/brief_overview.html
```diff
@@ -38,7 +38,28 @@
...
+        {% if messages %}
+            {% for category, message in messages %}
+                {% if category == 'error' %}
+                    <div class="banner-destructive-without-action">
```
idavidmcdonald: Is there not some sort of macro that means we don't need to write the HTML?

Example: https://github.com/alphagov/digitalmarketplace-brief-responses-frontend/blob/master/app/templates/briefs/view_response_result.html

app/templates/buyers/brief_overview.html
```diff
@@ -255,8 +276,8 @@ <h2 class="instruction-list-item-body">
       </ol>
     </div>
     <div class="column-one-third">
-      {% if brief.publishedAt %}
-        {% include 'buyers/_requirements_meta.html' %}
+      {% if brief.status == 'closed' and framework.status in ['live', 'expired'] %}
```
idavidmcdonald: You probably don't need to do the framework status check. If a brief has been published then the framework is either in a live or expired state (https://alphagov.github.io/digitalmarketplace-manual/framework-lifecycle-for-developers.html). Therefore this check is not actually doing anything.

tests/main/views/test_buyers.py
idavidmcdonald: I saw someone else structure a link assertion like this and now I've been copying it.
`assert doc.xpath("//a[text() = 'View the opportunity and its outcome'][contains(@href,'digital-outcomes-and-specialists/opportunities/1234')]")`

Feel free to use or not but it feels a it tighter to check link locations

tests/main/views/test_buyers.py
```diff
@@ -3676,3 +3642,122 @@ def test_award_brief_details_post_raises_400_and_displays_error_messages_and_pre
...
+                api_stubs.lot(slug='digital-outcomes', allows_brief=True),
+            ]
+        )
+        self.brief = api_stubs.brief(
+            user_id=123,
+            framework_slug='digital-outcomes-and-specialists-2',
+            lot_slug="digital-outcomes",
+            status='closed'
+        )['briefs']
+        self.brief["lotName"] = "Digital Outcomes"
```
idavidmcdonald: We definitely need to include this line?

tests/main/views/test_buyers.py
```diff
@@ -3676,3 +3642,122 @@ def test_award_brief_details_post_raises_400_and_displays_error_messages_and_pre
...
+        document = html.fromstring(res.get_data(as_text=True))
+        self.assert_breadcrumbs(res, extra_breadcrumbs=[
+            (
+                'I need a thing to do a thing',
+                '/buyers/frameworks/digital-outcomes-and-specialists-2/requirements/digital-outcomes/1234'
+            )
+        ])
+
+        page_title = document.xpath('//h1')[0].text_content()
+        assert "Why do you need to cancel" in page_title
```
idavidmcdonald: Should we be checking the brief title is placed in here?

tests/main/views/test_buyers.py
```diff
@@ -3676,3 +3642,122 @@ def test_award_brief_details_post_raises_400_and_displays_error_messages_and_pre
...
+
+        submit_button = document.xpath('//input[@class="button-save" and @value="Update requirement"]')
+        assert len(submit_button) == 1
+
+    def test_404_if_user_is_not_brief_owner(self):
+        self.data_api_client.get_brief.return_value['briefs']['users'][0]['id'] = 234
+
+        resp = self.client.get(self.url.format(brief_id=self.brief["id"]))
```
idavidmcdonald: I don't want to be tooo nitpicky but we tend to just call this `res` (appears to be convention across our entire codebase). Feel free to ignore me also

tests/main/views/test_buyers.py
```diff
@@ -3676,3 +3642,122 @@ def test_award_brief_details_post_raises_400_and_displays_error_messages_and_pre
...
+
+    @pytest.mark.parametrize('status', ['withdrawn', 'draft', 'live', 'cancelled', 'unsuccessful', 'awarded'])
+    def test_404_if_brief_not_closed(self, status):
+        self.data_api_client.get_brief.return_value['briefs']['status'] = status
+
+        resp = self.client.get(self.url.format(brief_id=self.brief["id"]))
+
+        assert resp.status_code == 404
+
+    @pytest.mark.parametrize('framework_status', ['coming', 'open', 'pending', 'standstill'])
```
idavidmcdonald: We should change this test to be the inverse. There is no way that a brief could be created if the framework is in any of the above states so it doesn't really make sense. However, we do care that we can access this page when the framework is live or withdrawn, so we should test that explicitly.

tests/main/views/test_buyers.py
```diff
@@ -3676,3 +3642,122 @@ def test_award_brief_details_post_raises_400_and_displays_error_messages_and_pre
...
+        resp = self.client.get(self.url.format(brief_id=123))
+
+        assert resp.status_code == 404
+
+    def test_that_no_option_chosen_triggers_error(self):
+        resp = self.client.post(self.url.format(brief_id=123))
+        document = html.fromstring(resp.get_data(as_text=True))
+        validation_message = document.xpath('//span[@class="validation-message"]')[0].text_content()
+        assert resp.status_code == 400
+        assert "Answer required" in validation_message
```
idavidmcdonald: Is it worth also asserting we are not calling the API to change the brief status here or am I being too cautious?

tests/main/views/test_buyers.py
idavidmcdonald: Yes please

app/main/forms/cancel.py
benvand: Sooo, basically we had a content loader version set up, working and ready to go. However, because design wants a single page question thing we couldn't pass the 'question' as the question attribute of the content loader YAML to the standard template because the standard template wouldn't allow for the bigger title. No problem, we just won't pass that from the YAML. We'll pass it as a heading.
Well apparently not because in order to pass tests every YAML question has to have 'question' attribute (seemingly arbitrarily).
By this point it got to feeling like we were trying to force a square peg through a round hole and it would be cleaner not to mess too heavily with the content loader and just WTForm it.

app/main/forms/cancel.py
benvand: That is used in some places. Notably in conjunction with the multi question - dependent question stuff. The outlier being the `company_summary.html` and the `ReuseDeclarationForm` which I created ages ago :|

'Answer required' is also used in some places but *drumroll please* the winner is, by a wide margin...
'You need to answer this question.' with 101+ matches in 100+ files!!! 🎉 

I'll get that changed.

tests/main/views/test_buyers.py
benvand: Ah, old habits die hard. Was a defacto convention in another codebase I was working on. I'll change it.

app/templates/buyers/brief_overview.html
benvand: In terms of the DRY I'm going to call it as out of scope of this ticket as we've got 8 other pretty much copy-paste occurrences of this pattern but I'll add a ticket to the tech debt board to get that cleaned up. Hopefully someone can find a nice solution and tie it into the base templates.

[messages-repetition-search.txt](https://github.com/alphagov/digitalmarketplace-briefs-frontend/files/1291712/messages-repetition-search.txt)

https://trello.com/c/fLiJa6tD/177-tidy-up-banner-display-code-and-allow-it-to-properly-accept-and-display-flash-messages

tests/main/views/test_buyers.py
benvand: 👍

app/templates/buyers/brief_overview.html
benvand: 👍

tests/main/views/test_buyers.py
benvand: This is quality.

tests/main/views/test_buyers.py
benvand: If the brief doesn't have a title attribute then it defaults back to lot name. The api stub doesn't have a 'title'.

However I can remove it it we do a `brief.get('lotName')` in the form instead of a `brief['lotName']`

tests/main/views/test_buyers.py
benvand: 👍

tests/main/views/test_buyers.py
benvand: Nah, I'm happy with adding that in.

tests/main/views/test_buyers.py
benvand: I think those tests do too much already.
If `test_show_awarded_brief_summary_page_for_live_and_expired_framework` fails I'd have no idea what was wrong.
If `test_cancel_link_present_on_closed_brief_summary` I'd know exactly what had happened.

app/main/forms/cancel.py
katstevens: Good to know. Worth documenting on the content loader README that it's not suitable for this sort of situation? (Or yet another tech debt ticket to introduce the ability to skip questions for tests if there's a header instead?)

idavidmcdonald: Other reasons why WTF is a good choice
- probably the quicker of the two ways to build
- The information that will be sent to the API is not a big dict of JSON data (content loader is good for this). Instead, it is just parsing one key and deciding which API endpoint to call
- We do not need to validate this using a JSON schema

app/main/views/buyers.py
idavidmcdonald: Cool

tests/main/views/test_buyers.py
idavidmcdonald: Cool, I don't mind either really. It was more it looked a little out of place.

tests/main/views/test_buyers.py
```diff
@@ -3676,3 +3645,125 @@ def test_award_brief_details_post_raises_400_and_displays_error_messages_and_pre
...
+        assert self.data_api_client.cancel_brief.called is False
+
+    def test_cancel_triggers_cancel_brief(self):
+        res = self.client.post(
+            self.url.format(brief_id=123), data={'cancel_reason': 'cancel'}
+        )
+
+        assert res.status_code == 302
+        self.data_api_client.cancel_brief.assert_called_once_with('123', user='buyer@email.com')
+        self.assert_flashes("updated-brief")
```
idavidmcdonald: It feels like this assertion should live in the test that checks the redirect as they are both to do with what should be shown to the user next

idavidmcdonald: Also ideally we would be asserting both the key and the value of the flash message we are setting

app/templates/buyers/brief_overview.html
```diff
@@ -38,8 +38,23 @@
...
         </div>
       {% endif %}
-   {% endblock %}
+
+      {% with messages = get_flashed_messages(with_categories=true) %}
+        {% if messages %}
+          {% for category, message in messages %}
+            {%
+            with
+            message = "You've updated {}".format(message.get('updated-brief')) if 'updated-brief' in message else message,
```
idavidmcdonald: Can we get a unit test for this please. You could likely do without the one for the error state but the positive state should be covered.

benvand: Done 👍

tests/main/views/test_buyers.py
benvand: Done 👍

PR #27, opened 2017-09-08
Dh 679 buyer account page

DilwoarH:
Ticket:  https://trello.com/c/FiR3NLyZ/679-buyer-account-page

### DIFF
--- a/tests/main/views/test_buyers.py
+++ b/tests/main/views/test_buyers.py
@@ -3720,3 +3720,23 @@ class TestCancelBrief(BaseApplicationTest):
         res = self.client.get(expected_url)
         flash_div = html.fromstring(res.get_data(as_text=True)).xpath('//div[@class="banner-success-without-action"]')
         assert flash_div[0].text_content().strip() == "You've updated 'My Amazing Brief'"
+
+
+class TestBuyerAccountOverview(BaseApplicationTest):
+
+    def setup_method(self, method):
+        super(TestBuyerAccountOverview, self).setup_method(method)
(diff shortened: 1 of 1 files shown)


DilwoarH:
cherry picked onto another PR

tests/main/views/test_buyers.py
```diff
@@ -3720,3 +3720,23 @@ def test_flash_message_shown_on_brief_cancel(self):
         res = self.client.get(expected_url)
         flash_div = html.fromstring(res.get_data(as_text=True)).xpath('//div[@class="banner-success-without-action"]')
         assert flash_div[0].text_content().strip() == "You've updated 'My Amazing Brief'"
+
+
+class TestBuyerAccountOverview(BaseApplicationTest):
+
+    def setup_method(self, method):
+        super(TestBuyerAccountOverview, self).setup_method(method)
```
benvand: This test breaks every test after it that uses a `data_api_client_patch`.
```
    def teardown_method(self, method):
        super(TestBuyerAccountOverview, self).teardown_method(method)
```
We need to stop the patch at the end of the test.


Also why can't this go in the `TestBuyerDashboard` class above?

benvand: > This test breaks every test after it that uses a data_api_client_patch
This would have been hard to spot in all fairness but this test still doesn't need its own class.

DilwoarH: looks like you've added in the changes already, will keep this in mind for future.

PR #28, opened 2017-09-11
Flake8 for Briefs front end

katstevens:
Replaces pep8. Based on the flake8 config for the API repo (ignoring import errors in `__init__` files etc). 

Tidying included:
- whitespace, unused imports/vars etc
- Avoiding using the `unicode` built-in (so we can support both Py2/Py3)
- Removed a duplicate test
- Removed a duplicate dict key in a test
- Renamed a duplicate error endpoint (410 is 'gone', 404 is 'not found')

I ignored complexity errors, as we haven't addressed these in other repos as far as I know.

`make test` passes in both Py2 and Py3.


app/main/views/buyers.py
```diff
@@ -533,8 +533,11 @@ def get_context_data(self, **kwargs):
         if brief['status'] not in CLOSED_PUBLISHED_BRIEF_STATUSES:
             abort(404)
 
-        text_type = str if sys.version_info[0] == 3 else unicode
-        filename = inflection.parameterize(text_type(brief['title']))
+        # Filename must be unicode
+        if sys.version_info[0] == 3:
```
galund: Don't think we need this anymore at all now we're not supporting python 2.x?

app/main/views/buyers.py
```diff
@@ -533,8 +533,11 @@ def get_context_data(self, **kwargs):
         if brief['status'] not in CLOSED_PUBLISHED_BRIEF_STATUSES:
             abort(404)
 
-        text_type = str if sys.version_info[0] == 3 else unicode
-        filename = inflection.parameterize(text_type(brief['title']))
+        # Filename must be unicode
```
benvand: http://pythonhosted.org/six/#six.text_type

Six can take care of this and we've got it as a dependency.

tests/main/helpers/test_ods.py
```diff
@@ -1,11 +1,10 @@
 import mock
-import unittest
 import functools
 
 import app.main.helpers.ods as ods
 
 from hypothesis import strategies as st
-from hypothesis import given, example, assume
+from hypothesis import given, example
```
benvand: I'm pretty _partial_ to removing this pattern whenever I find it.
LOOOOOL

But not in this PR cause it doesn't cause a flake8 error on its own. No matter how much I wish it did.

benvand:
👍  Happy for this to go in as soon as we've made a decision on the py3 thing

app/main/views/buyers.py
katstevens: If we're happy to drop Py2 here then great!

app/main/views/buyers.py
katstevens: Sweet, I didn't know we had that - much neater all round.

app/main/views/buyers.py
```diff
@@ -28,6 +27,7 @@
 from odf.style import TextProperties, TableRowProperties, TableColumnProperties, TableCellProperties, FontFace
 
 from io import BytesIO
+from six import text_type
```
galund: I still maintain we don't even need this, now we're definitively (*except for scripts) a P3K shop. @allait will comment if I misunderstood something :)

benvand: @galund In terms of this being a purely flake8 PR I'd say allow this in order to change as little functionality as possible.
I'd like to see someone go through all the repos and remove _all_ the py2-3 compat stuff at once and we can officially drop py2.

galund: FWIW I think we should be doing that as we go, I can't imagine a product person prioritising a ticket that said "go through all the repos" doing _anything_ much ;-)

katstevens: Well I probably do have a spare hour today 😸

benvand: @galund sure we _should_ but we wont. We did get tickets in to flake8 all the repos to be fair. Just needs someone pushing hard enough.

benvand: Firebreak sprint? ;)

katstevens: I think it's reasonable to have 'drop Py2 support' in a separate PR, so we have more control over when we release it?

galund: Absolutely, doesn't have to be here, up to you. But we already dropped support, so this is just about removing redundant code, as and when it's convenient.

katstevens:
In the end I decided to remove the Py2 support here - we no longer run apps in Py2 and hopefully this will force anyone still using local Py2 venvs to update them!

PR #29, opened 2017-09-12
Brief overview page for Live and Draft briefs

katstevens:
Trello ticket: https://trello.com/c/k6d0ggj0/835-1-update-brief-overview-page-based-on-draft-open-brief-status

- Changed copy and link locations according to prototype
- Aligned the links with the top of the instructions list (as per Laurence's recommendation)
- Removed a CSS class that was no longer being used (as we've got rid of `#requirements-meta`), and a test that only checked that the div wasn't being shown.


app/templates/buyers/brief_overview.html
```diff
@@ -272,16 +272,13 @@ <h2 class="instruction-list-item-body">
     <div class="column-one-third">
       {% if brief.status == 'closed' %}
         <a href="{{ url_for('buyers.cancel_brief', framework_slug=framework.slug, lot_slug=brief.lotSlug, brief_id=brief.id) }}">Cancel requirement</a>
-     {% endif %}
+      {% elif brief.status == 'live' %}
+        <a href="https://www.gov.uk/guidance/how-to-make-changes-to-your-published-digital-outcomes-and-specialists-requirements#when-to-withdraw-your-requirements">Withdraw requirement</a>
+      {% elif framework.status == 'live' and brief.status == 'draft' %}
```
benvand: Do we care if the framework is live or expired?

tests/main/views/test_buyers.py
```diff
@@ -1693,7 +1693,8 @@ def test_show_draft_brief_summary_page(self, data_api_client):
             ]
 
             assert "Awarded to " not in page_html
-            assert document.xpath('//a[contains(text(), "Delete")]')
+            assert document.xpath('//a[contains(text(), "Delete draft requirement")]')
+            assert not document.xpath('//a[contains(text(), "Withdraw requirement")]')
```
benvand: We don;t need this `assert not` here. It should either be important enough to be a dedicated test or we shouldn't have it at all. `test_show_draft_brief_summary_page` is about things that _should_ be there.

benvand:
Not a fan of asserting nots. Where does it stop is the problem. If we want to have tests that check only the correct links are shown then that's cool but by asserting not in there 
a) we're blurring the lines of what those tests are there to check (are they checking that what's should be there is there or what shouldn't be there isn't there?)
b) we're making the tests less useful by making them bigger
````
@pytest.mark.parametrize('expected_link,status'   ... ... )
def test_show_correct_links_on_brief_overview(self, staus, expected_link):
    ...
```

If the above fails then I know exactly what's up but if `test_show_draft_brief_summary_page` fails it could be anything on that entire page.


I've checked the links though and they look good, completely as per content 👍

app/templates/buyers/brief_overview.html
katstevens: Not sure! Is there any problem deleting draft requirements on an expired framework?

tests/main/views/test_buyers.py
katstevens: I was trying to be consistent with the assertions across all the status tests 😸 Agreed it's not ideal to assert the absence of something. How about instead we assert the exact list of all the links in that side div - there should only be one `column-one-third` div per overview page.

katstevens:
Another thought (probably outside the scope of this PR!): where should we draw the line between asserting content in unit tests and in functional tests?

app/templates/buyers/brief_overview.html
benvand: No, suppliers can't apply for them at that point and because there will only be a max 2 week window for something to be draft on an expired framework before it becomes live..

katstevens: Cool - will remove.

tests/main/views/test_buyers.py
```diff
@@ -1653,6 +1653,17 @@ def test_404_if_brief_does_not_belong_to_user(self, data_api_client):
 
 @mock.patch('app.main.views.buyers.data_api_client', autospec=True)
 class TestBriefSummaryPage(BaseApplicationTest):
+
```
benvand: Not a blocker but I think these tests could be a bit more readable with something like this

```
SIDE_LINKS_XPATH = '//div[@class="column-one-third"]//a'
INSTRUCTION_LINKS_XPATH = '//main[@id="content"]//ul/li/a'

@staticmethod
def _get_links(document, xpath, text_only=False):
    if text_only:
        return [e.text_content() for e in document.xpath('//main[@id="content"]//ul/li/a')]
    else:
        return [(e.text_content(), e.get('href')) for e in document.xpath('//div[@class="column-one-third"]//a')]

def test_foo(self, dataapiclient):
    ...
    assert self._get_links(document, self.SIDE_LINKS_XPATH) == [('Foo', 'my/cool/foo/url')]
```

The above assertion tells us that we're getting the side links from the document using the xpath and that they should match the list we've defined. We don't really need to go and look at what `_get_links` does because it's super obvious from the name and signature. I prefer not to extrapolate too much out in tests because it breaks the flow of reading a test if you have to keep jumping around to different methods.

This is personal preference though. Approving now.

katstevens: I like the idea of making it clear we're using xpath, thanks! Also one generic helper is always better than two 😸

PR #31, opened 2017-09-14
818 add no journey to the award flow

benvand:
https://trello.com/c/LcbXbFx7/818-3-add-no-journey-to-the-award-flow

This PR joins together the award flow and cancel flow with an 'award or cancel' step:

award_or_cancel_flow

It adds a new url to the cancel view '/award-cancel'. When accessing the view through this url the form text on the page is updated to read 'Why didn’t you award a contract for [brief name]?' instead of the default 'Why do you need to cancel [brief name]?' detailed here: https://docs.google.com/document/d/1n-9ucQVguGvp6za2vr_VN0nciZOdysKPHDMESCkUkZs/edit#heading=h.l4xho8gj18u3
This comes with associated tweaks to the cancel brief form and view.

It also adds a completely new form, view and template (https://award-prototype.cloudapps.digital/award/status/award?opportunity_id=1):
newform

The view has 4 'routes'. GET shows the form, POST with 'yes' takes you to the form to pick a supplier you awarded your brief to, POST with 'no' takes you to the cancel flow through the new url asking 'why you didn't award',  POST with 'back' returns you to the DOS requirements overview.

There is quite a bit or repetition in the form files and the view and test files. Refactoring this in this PR would obscure the new logic and tests and is somewhat out of scope of this PR. I've raised a tech debt ticket to addres this at a future date:
https://trello.com/c/xTjixqZX/195-reduce-repetition-of-code-in-app-main-forms-app-main-views-tests-main-testviews-in-briefs-frontend

### DIFF
--- a/app/main/forms/award_or_cancel.py
+++ b/app/main/forms/award_or_cancel.py
@@ -1,0 +1,22 @@
+from flask.ext.wtf import Form
+from wtforms import RadioField, validators
+
+
+class AwardOrCancelBriefForm(Form):
+    """Form for the buyer to tell us whether they want to award or cancel the requirement
+    """
+    choices = [
+        ('yes', 'Yes'),
+        ('no', 'No'),
+        ('back', 'We are still evaluating suppliers'),
+    ]
+    award_or_cancel_decision = RadioField(
+        validators=[validators.InputRequired(message='You need to answer this question.')],
+        choices=choices
+    )
+
+    def __init__(self, brief, *args, **kwargs):
+        super(AwardOrCancelBriefForm, self).__init__(*args, **kwargs)
+        brief_name = brief.get('title', brief.get('lotName', ''))
+        self.award_or_cancel_decision.label.text = "Have you awarded a contract for {}?".format(brief_name)
+        self.award_or_cancel_decision.toolkit_macro_options = [{'value': i[0], 'label': i[1]} for i in self.choices]
--- a/app/main/views/buyers.py
+++ b/app/main/views/buyers.py
@@ -495,11 +495,22 @@ def award_or_cancel_brief(framework_slug, lot_slug, brief_id):
                 # We should never get here as the form validates the answers against the available choices.
                 abort(500, "Unexpected answer to award or cancel brief")
 
+    breadcrumbs = get_briefs_breadcrumbs([{
+        "label": brief['title'],
+        "link": url_for(
+            ".view_brief_overview",
+            framework_slug = brief['frameworkSlug'],
+            lot_slug = brief['lotSlug'],
+            brief_id = brief['id']
+        )
+    }])
+
     return render_template(
         "buyers/award_or_cancel_brief.html",
         brief=brief,
         form=form or AwardOrCancelBriefForm(brief),
         errors=errors,
+        breadcrumbs=breadcrumbs
     ), 200 if not errors else 400
 
 
--- a/app/main/views/buyers.py
+++ b/app/main/views/buyers.py
@@ -600,8 +600,8 @@ def award_brief(framework_slug, lot_slug, brief_id):
 def cancel_brief(framework_slug, lot_slug, brief_id):
     form = None
     errors = {}
-    end_point = request.endpoint.strip(request.blueprint + '.')
-    label_text = "Why didn't you award a contract for {}?" if end_point == 'cancel_award_brief' else None
+    award_flow = request.endpoint.strip(request.blueprint + '.') == 'cancel_award_brief'
+
     get_framework_and_lot(
         framework_slug,
         lot_slug,
@@ -678,6 +678,7 @@ def cancel_brief(framework_slug, lot_slug, brief_id):
         form=form or CancelBriefForm(brief, label_text),
         errors=errors,
         breadcrumbs=breadcrumbs,
+        previous_page_url=previous_page_url
     ), 200 if not errors else 400
 
 
--- a/app/main/views/buyers.py
+++ b/app/main/views/buyers.py
@@ -545,7 +545,7 @@ def award_brief(framework_slug, lot_slug, brief_id):
         )
 
     if request.method == "POST":
-        form = AwardedBriefResponseForm(request.form, brief_responses=brief_responses)
+        form = AwardedBriefResponseForm(brief_responses, request.form)
         if not form.validate_on_submit():
             form_errors = [{'question': form[key].label.text, 'input_name': key} for key in form.errors]
             return render_template(
@@ -576,7 +576,7 @@ def award_brief(framework_slug, lot_slug, brief_id):
                 )
             )
 
-    form = AwardedBriefResponseForm(brief_responses=brief_responses)
+    form = AwardedBriefResponseForm(brief_responses)
     pending_brief_responses = list(filter(lambda x: x.get('awardDetails', {}).get('pending'), brief_responses))
     form['brief_response'].data = pending_brief_responses[0]["id"] if pending_brief_responses else None
 
@@ -588,10 +588,20 @@ def award_brief(framework_slug, lot_slug, brief_id):
     ), 200
 
 
-@main.route('/frameworks/<framework_slug>/requirements/<lot_slug>/<brief_id>/cancel', methods=['GET', 'POST'])
+@main.route(
+    '/frameworks/<framework_slug>/requirements/<lot_slug>/<brief_id>/cancel',
+    methods=['GET', 'POST'],
+)
+@main.route(
+    '/frameworks/<framework_slug>/requirements/<lot_slug>/<brief_id>/cancel-award',
+    methods=['GET', 'POST'],
+    endpoint="cancel_award_brief"
+)
 def cancel_brief(framework_slug, lot_slug, brief_id):
     form = None
     errors = {}
+    end_point = request.endpoint.strip(request.blueprint + '.')
+    label_text = "Why didn't you award a contract for {}?" if end_point == 'cancel_award_brief' else None
     get_framework_and_lot(
         framework_slug,
         lot_slug,
@@ -605,8 +615,25 @@ def cancel_brief(framework_slug, lot_slug, brief_id):
     if brief["status"] != "closed":
         abort(404)
 
+    if award_flow:
+        label_text = "Why didn't you award a contract for {}?"
+        previous_page_url = url_for(
+            '.award_or_cancel_brief',
+            framework_slug=brief['frameworkSlug'],
+            lot_slug=brief['lotSlug'],
+            brief_id=brief['id']
+        )
+    else:
+        # Use default label text
+        label_text = 'Why do you need to cancel {}?'
+        previous_page_url = url_for(
+            '.view_brief_overview',
+            framework_slug=brief['frameworkSlug'],
+            lot_slug=brief['lotSlug'],
+            brief_id=brief['id']
+        )
     if request.method == "POST":
-        form = CancelBriefForm(request.form, brief=brief)
+        form = CancelBriefForm(brief, label_text, request.form)
         if not form.validate_on_submit():
             errors = {
                 key: {'question': form[key].label.text, 'input_name': key, 'message': form[key].errors[0]}
@@ -648,7 +675,7 @@ def cancel_brief(framework_slug, lot_slug, brief_id):
     return render_template(
         "buyers/cancel_brief.html",
         brief=brief,
-        form=form or CancelBriefForm(brief=brief),
+        form=form or CancelBriefForm(brief, label_text),
         errors=errors,
         breadcrumbs=breadcrumbs,
     ), 200 if not errors else 400
--- a/app/main/views/buyers.py
+++ b/app/main/views/buyers.py
@@ -19,6 +19,8 @@ from ..helpers.buyers_helpers import (
 from ..helpers.ods import SpreadSheet
 from ..forms.awards import AwardedBriefResponseForm
 from ..forms.cancel import CancelBriefForm
+from ..forms.award_or_cancel import AwardOrCancelBriefForm
+
 
 from dmapiclient import HTTPError
 from dmutils.dates import get_publishing_dates
@@ -449,6 +451,56 @@ def view_brief_responses(framework_slug, lot_slug, brief_id):
         brief=brief,
         breadcrumbs=breadcrumbs
     ), 200
+
+
+@main.route('/frameworks/<framework_slug>/requirements/<lot_slug>/<brief_id>/award', methods=['GET', 'POST'])
+def award_or_cancel_brief(framework_slug, lot_slug, brief_id):
+    form = None
+    errors = {}
+    get_framework_and_lot(
+        framework_slug,
+        lot_slug,
+        data_api_client,
+        allowed_statuses=['live', 'expired'],
+        must_allow_brief=True,
+    )
+    brief = data_api_client.get_brief(brief_id)["briefs"]
+
+    if not is_brief_correct(brief, framework_slug, lot_slug, current_user.id):
+        abort(404)
+
+    if brief['status'] != "closed":
+        abort(404)
+
+    if request.method == "POST":
+        form = AwardOrCancelBriefForm(brief, request.form)
+        if not form.validate_on_submit():
+            errors = {
+                key: {'question': form[key].label.text, 'input_name': key, 'message': form[key].errors[0]}
+                for key, value in form.errors.items()
+            }
+        else:
+            answer = form.data.get('award_or_cancel_decision')
+            if answer == 'back':
+                return redirect(url_for('.buyer_dos_requirements'))
+            elif answer == 'yes':
+                return redirect(
+                    url_for('.award_brief', framework_slug=framework_slug, lot_slug=lot_slug, brief_id=brief_id)
+                )
+            elif answer == 'no':
+                return redirect(url_for(
+                    '.cancel_award_brief', framework_slug=framework_slug, lot_slug=lot_slug, brief_id=brief_id)
+                )
+            else:
+                # We should never get here as the form validates the answers against the available choices.
+                abort(500, "Unexpected answer to award or cancel brief")
+
+    return render_template(
+        "buyers/award_or_cancel_brief.html",
+        brief=brief,
+        form=form or AwardOrCancelBriefForm(brief),
+        errors=errors,
+    ), 200 if not errors else 400
 
 
 @main.route('/frameworks/<framework_slug>/requirements/<lot_slug>/<brief_id>/award-contract', methods=['GET', 'POST'])
--- a/app/templates/buyers/award_details.html
+++ b/app/templates/buyers/award_details.html
@@ -52,7 +52,7 @@
       {%
           with
           url=url_for(".award_brief", framework_slug=brief.frameworkSlug, lot_slug=brief.lotSlug, brief_id=brief.id),
-          text="Back to previous page"
+          text="Previous page"
       %}
           {% include "toolkit/secondary-action-link.html" %}
       {% endwith %}
--- a/app/templates/buyers/award_or_cancel_brief.html
+++ b/app/templates/buyers/award_or_cancel_brief.html
@@ -1,0 +1,75 @@
+{% extends "_base_page.html" %}
+
+{% import "toolkit/forms/macros/forms.html" as forms %}
+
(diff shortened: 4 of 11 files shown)


app/templates/buyers/award_details.html
```diff
@@ -65,7 +65,7 @@
       {%
           with
           url=url_for(".award_brief", framework_slug=brief.frameworkSlug, lot_slug=brief.lotSlug, brief_id=brief.id),
-          text="Back to previous page"
+          text="Previous page"
       %}
```
benvand: https://award-prototype.cloudapps.digital/award/suppliers/award

tests/main/views/test_buyers.py
```diff
@@ -3617,6 +3617,43 @@ def test_cancel_brief_200s_with_correct_default_content(self):
...
 
+        expected_previous_page_link_text = 'Previous page'
+        expected_previous_page_link_url = (
+            '/buyers/frameworks/digital-outcomes-and-specialists-2/requirements/digital-outcomes/1234'
+        )
+
+        assert (
+            document.xpath("//a[text()='{}']/@href".format(expected_previous_page_link_text))[0] ==
+            expected_previous_page_link_url
+        )
```
katstevens: Hmm... rather than searching for the expected link text, could we instead assert that there is a single link inside the `secondary-action-link` div (we should only have one of those per page, right?) and that this link has the expected text/url? (Same for the test below)

app/main/forms/award_or_cancel.py
```diff
@@ -0,0 +1,23 @@
+from flask.ext.wtf import Form
+from wtforms import RadioField, validators
+
+
+class AwardOrCancelBriefForm(Form):
+    """Form for the buyer to tell us why they cancel the requirement
```
katstevens: I think this docstring needs tweaking

tests/main/views/test_buyers.py
```diff
@@ -3598,7 +3598,7 @@ def teardown_method(self, method):
         self.data_api_client_patch.stop()
         super(TestCancelBrief, self).teardown_method(method)
 
-    def test_cancel_brief_200s_with_correct_default_content(self):
+    def test_cancel_brief_form_displays_correctly(self):
```
katstevens: Can we adjust this test name to something like `test_cancel_brief_form_displays_correctly_accessed_from_brief_overview_link` to be consistent with the one below?

tests/main/views/test_buyers.py
```diff
@@ -3700,3 +3737,110 @@ def test_flash_message_shown_on_brief_cancel(self):
...
+
+    def test_that_no_option_chosen_triggers_error(self):
+        res = self.client.post(self.url.format(brief_id=123))
+
+        document = html.fromstring(res.get_data(as_text=True))
+        validation_message = document.xpath('//span[@class="validation-message"]')[0].text_content()
+
+        assert res.status_code == 400
+        assert "You need to answer this question." in validation_message
+        assert self.data_api_client.cancel_brief.called is False
```
katstevens: I'm not 100% sure this assertion is useful? Do we ever cancel via the API in this view? (Also if it turns out we do need it, I'd rather have the assertion be `is None`!)

tests/main/views/test_buyers.py
```diff
@@ -3700,3 +3737,110 @@ def test_flash_message_shown_on_brief_cancel(self):
...
+
+    def test_yes_redirect(self):
+        self.login_as_buyer()
+        res = self.client.post(self.url.format(brief_id=self.brief['id']), data={'award_or_cancel_decision': 'yes'})
+
+        redirect_text = html.fromstring(res.get_data(as_text=True)).text_content().strip()
+        expected_url = (
+            '/buyers/frameworks/digital-outcomes-and-specialists-2/requirements/digital-outcomes/{}/award-contract'
+        ).format(self.brief['id'])
```
katstevens: I may have misunderstood this, but are we asserting here (and in the other redirect tests) that the redirect url appears somewhere in the page content of destination page? Does that then follow that we have actually been redirected to that url?

Elsewhere in this test module we're able to assert redirect locations like this, without needing to do any text wrangling:
```
res = self.client.post('/wherever')
assert res.status_code == 302
assert res.location == 'http://localhost/destination'
```

app/templates/buyers/dashboard.html
```diff
@@ -127,7 +127,7 @@
             <div class="padding-bottom-small">
                 <p><a href="{{ url_for('.view_brief_responses', framework_slug=item.frameworkSlug, lot_slug=item.lot, brief_id=item.id) }}">View responses</a></p>
                 {% if item.status == "closed" %}
-                    <p><a class="award-contract-link" href="{{ url_for('.award_brief', framework_slug=item.frameworkSlug, lot_slug=item.lot, brief_id=item.id) }}">Tell us who won this contract</a></p>
+                    <p><a class="award-contract-link" href="{{ url_for('.award_or_cancel_brief', framework_slug=item.frameworkSlug, lot_slug=item.lot, brief_id=item.id) }}">Let suppliers know the outcome</a></p>
```
katstevens: FYI this copy change will mean the functional tests (in `award_requirements.feature`) will fail on preview/staging - the link text and URL are both hardcoded there.

app/main/forms/award_or_cancel.py
benvand: Doh! I've got one unpushed commit locally. Guess what it does :grin:

app/main/views/buyers.py
```diff
@@ -533,6 +533,22 @@ def cancel_brief(framework_slug, lot_slug, brief_id):
...
+    if award_flow:
+        label_text = "Why didn't you award a contract for {}?"
+        previous_page_url = url_for(
+            '.award_or_cancel_brief',
+            framework_slug=brief['frameworkSlug'],
+            lot_slug=brief['lotSlug'],
+            brief_id=brief['id']
+        )
+    else:
+        label_text = None
```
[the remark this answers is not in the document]
benvand: I thought about this. I wanted the delete form to have a default option and if it has a default option then doing it here would be dupplicating.

I could make the delete form default something really generic like 'Do you want to delete?' but then there's no need for the briefs kwarg.
Unless I create a delete form base that is only used as a subclass for the 2 'personalised' delete brief forms that accept brief?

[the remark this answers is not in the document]
katstevens: Hmm, tricky. How about making the label_text an optional kwarg on the form? So we only overwrite the default if it's supplied? Something like

`label_text = kwargs.get('label_text', "Why do you need to cancel {}?")` ?

tests/main/views/test_buyers.py
benvand: Hmmm maybe I should move it ito another test called:
`test_that_no_option_chosen_does_not_trigger_update` ?

The comparison to False is a funny one. `.called` on a mock will always return either the [python `True` or python `False`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.called) object depending on whether the mock function was called. Because there are only ever one of each of these. and because the `is` keyword ensures that 2 objects have the same `id` `is` is the right comparison to use here.
```
In [1]: [] is []
Out[1]: False

In [1]: [] is False
Out[1]: False

In [2]: False is False
Out[2]: True
```

We can of course just use `assert not mock.called`. But because `0, (), [], {}, None, False, ''` are all `not True`, and because we know what we should be getting `is False` is the one.

On the other hand we can ignore all of the above because I should be using [`mock.assert_not_called()`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_not_called) apparently.

tests/main/views/test_buyers.py
benvand: Nice catch

tests/main/views/test_buyers.py
```diff
@@ -3700,3 +3717,110 @@ def test_flash_message_shown_on_brief_cancel(self):
...
+
+        redirect_text = html.fromstring(res.get_data(as_text=True)).text_content().strip()
+        expected_url = (
+            '/buyers/frameworks/digital-outcomes-and-specialists-2/requirements/digital-outcomes/{}/cancel-award'
+        ).format(self.brief['id'])
+
+        assert res.status_code == 302
+        assert expected_url in redirect_text
+
+    def test_back_redirect(self):
```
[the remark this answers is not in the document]
benvand: :+1: will do

tests/main/views/test_buyers.py
benvand: So I did this because I don't want changes in the frontend toolkit or refactoring of scss class names to fail our tests. 
If I managed to put the `Previous page` link in the correct position on the page _without_ using `secondary-action-link` for whatever reason the requirements of the ticket would still be satisfied and the tests should still pass.
If you feel strongly I'll change it.

katstevens: Yeah, that makes sense, happy for it to stay then.

tests/main/views/test_buyers.py
katstevens: 😸 Yes I'd much prefer `assert_not_called` if we were going to keep it. Can we do the same for `test_400_if_incorrect_status_supplied`?

app/main/views/buyers.py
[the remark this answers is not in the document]
benvand: So I've got something like that in the form:
`label_text = kwargs.pop('label_text') or 'Why do you need to cancel {}?'`
which I'm changing to the more pythonic:
`label_text = kwargs.pop('label_text', 'Why do you need to cancel {}?')`

But I have to define `label_text` here because it's used in 3 places in the view. I've put a comment now about the default text in the view.

What you think?

tests/main/views/test_buyers.py
benvand: Done :+1:

tests/main/views/test_buyers.py
benvand: Assert not called and broken out into different tests :+1:

tests/main/views/test_buyers.py
benvand: Done :+1:

tests/main/views/test_buyers.py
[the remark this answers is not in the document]
benvand: Done :+1:

app/main/views/buyers.py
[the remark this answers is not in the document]
benvand: `label_text = kwargs.pop('label_text', 'Why do you need to cancel {}?')` doesn't work. It'll return None if the key exists regardless of the value.
I expected 
```
{'foo': None}.pop('foo', 'bar') == 'bar'
```
I got
```
{'foo': None}.pop('foo', 'bar') == None
```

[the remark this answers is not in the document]
katstevens: Any reason why we can't do `.get` on the kwargs instead (I'd hope that would return `bar`)?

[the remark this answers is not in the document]
benvand: Just figured out what you're getting at. I'll admit it did feel a bit weird.

We instinctively wanted to preserve the signature of the underlying base class by passing `form` as the first argument to our forms. 
The problem was that in doing so we couldn't use kwargs for _our_ arguments because they would be mistaken for positional kwargs in the **kwargs when passed to the base classes __init__ method, unless we popped them out of the kwargs dict.

```
# This is our base class. It takes a form_data kwarg
class Form(object):
    def __init__(self, form_data='something'):
        print form_data    
        

# First thought:
class OurForm(Form):
    def __init__(self, brief_responses, *args, **kwargs):
        super(OurForm, self).__init__(*args, **kwargs)
        print 'do something with our {}'.format(brief_responses)

OurForm('our_brief_responses', 'our_form_data')
# > our_form_data
# > do something with our our_brief_responses

# But that ^ doesn't let us pass form in first like the parent class, to remedy that let's try:

class OurForm(Form):
    def __init__(self, *args, **kwargs):
        kwargs.get('brief_responses')
        super(OurForm, self).__init__(*args, **kwargs)
        print 'do something with our {}'.format(brief_responses) 

# > OurForm('our_form_data', brief_responses='our_brief_responses')
# > TypeError: __init__() got an unexpected keyword argument 'brief_responses'


# The base class doesn't like that kwarg
# OK TypeError, we'll just pop our kwarg and not pass it on through
class OurForm(Form):
    def __init__(self, *args, **kwargs):
        brief_responses = kwargs.pop('brief_responses')
        super(OurForm, self).__init__(*args, **kwargs)
        print 'do something with our {}'.format(brief_responses) 

OurForm('our_form_data', brief_responses='our_brief_responses')
# > our_form_data
# > do something with our our_brief_responses


# OK that works (This is where the AwardForm was at) but...
# The plot thickens, default arguments
class OurForm(Form):
    def __init__(self, brief_responses='some default brief_responses', *args, **kwargs):
        super(OurForm, self).__init__(*args, **kwargs)
        print 'do something with our {}'.format(brief_responses) 

OurForm('our_form_data', brief_responses='our_actual_brief_resposnes')
# > TypeError: __init__() got multiple values for keyword argument 'brief_responses'

# Another type error? Wait, it thinks our_form_data is brief_responses too because of the positional args!


# We also tried `pop` and `get` without `or` for the default:
class OurForm(Form):
    def __init__(self, *args, **kwargs):
        brief_responses = kwargs.pop('brief_responses', 'some default brief_responses')
        super(OurForm, self).__init__(*args, **kwargs)
        print 'do something with our {}'.format(brief_responses)

OurForm('our_form_data', brief_responses='our_brief_responses')
> # do something with our our_brief_responses
> # our_form_data

OurForm('our_form_data', brief_responses=None)
> # do something with our None
> # our_form_data

# Pop didn't work ^ because it just looks for the key in the dict, and if its value is None then so be it


# Hence 'do something with our None'
# Let's try get:
class OurForm(Form):
    def __init__(self, *args, **kwargs):
        brief_responses = kwargs.get('brief_responses', 'some default brief_responses')
        super(OurForm, self).__init__(*args, **kwargs)
        print 'do something with our {}'.format(brief_responses)

OurForm('our_form_data', brief_responses='our_brief_responses')
# > TypeError: __init__() got an unexpected keyword argument 'brief_responses'
OurForm('our_form_data', brief_responses=None)
# > TypeError: __init__() got an unexpected keyword argument 'brief_responses'

# Get didn't work ^ because it doesn't remove the kwarg, and in any case it behaves the same as pop 
# just looks for the key in the dict, and if its value is None then so be it


# The solution I had in my PR was to use `or` to compare what we were getting from popping the kwarg:
class OurForm(Form):
    def __init__(self, *args, **kwargs):
        brief_responses = kwargs.pop('brief_responses') or 'some default brief_responses'
        super(OurForm, self).__init__(*args, **kwargs)
        print 'do something with our {}'.format(brief_responses)

OurForm('our_form_data', brief_responses='our_brief_responses')
# > our_form_data
# > do something with our our_brief_responses
OurForm('our_form_data', brief_responses=None)
# > our_form_data
# > do something with our some default brief_responses

# Which worked but felt a little kludgy, why do we need to pop at all..? The answer is we don't:


# As long as we stop trying to preserve the signature of the base class and pass our _required_ arguments first
# we can circle back round to our first thought. No TypeErrors, no duplicate kwargs. We never touch form_data
# The form doesn't need a default in a kwarg for brief_responses or label_text, it can legitimately require those
class OurForm(Form):
    def __init__(self, brief_responses, *args, **kwargs):
        super(OurForm, self).__init__(*args, **kwargs)
        print 'do something with our {}'.format(brief_responses)

OurForm('our_brief_responses', 'our_form_data')
# > our_form_data
# > do something with our our_brief_responses

# Works!
# There we have it,, that was the thought processes and attempts behind crating the AwardOrCancelForm, CancelForm and my refactor of the AwardForm
# Thanks for pushing me to the conclusion! Sorry for the essay! I wanted to write it down for myself too.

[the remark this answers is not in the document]
Comment (katstevens) [reply app/main/views/buyers.py]: Awesome - thanks for the explanation, all makes sense 👍

File: app/templates/buyers/award_details.html (hunk 1 above)
Comment (idavidmcdonald) [reply app/templates/buyers/award_details.html]: The prototype is inconsistent with what we are already doing on the marketplace which is using 'Back to previous page'.

It was previously agreed, probably on a different story, that we go with consistency for the moment.

Saying that, this is an app for buyers so they probably won't have seen the equivalent text in the brief responses app and so wouldn't actually see the inconsistencies.

Something to clarify with content just in case here.

File: app/templates/buyers/award_or_cancel_brief.html [hunk 10]
```diff
@@ -0,0 +1,75 @@
...
+{% block breadcrumb %}
+  {%
+    with items = [
+      {
+        "link": "/",
+        "label": "Digital Marketplace"
+      },
+      {
+        "link": url_for(".buyer_dashboard"),
+        "label": "Your account"
```
idavidmcdonald: If the compare and decide team have merged their breadcrumbs work, we should build on top of the functionality they've added as otherwise when they remove the feature flag we might be missing a breadcrumb here

app/main/views/buyers.py
```diff
@@ -382,6 +384,56 @@ def view_brief_responses(framework_slug, lot_slug, brief_id):
...
+        allowed_statuses=['live', 'expired'],
+        must_allow_brief=True,
+    )
+    brief = data_api_client.get_brief(brief_id)["briefs"]
+
+    if not is_brief_correct(brief, framework_slug, lot_slug, current_user.id):
+        abort(404)
+
+    if brief['status'] != "closed":
+        abort(404)
```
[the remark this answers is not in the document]
idavidmcdonald: It might not just be this view (the other form views too)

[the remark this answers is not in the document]
katstevens: I've looked at this as part of https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/36, but it will need changing once this is merged obviously.

app/main/views/buyers.py
```diff
@@ -451,6 +453,56 @@ def view_brief_responses(framework_slug, lot_slug, brief_id):
...
+        form = AwardOrCancelBriefForm(brief, request.form)
+        if not form.validate_on_submit():
+            errors = {
+                key: {'question': form[key].label.text, 'input_name': key, 'message': form[key].errors[0]}
+                for key, value in form.errors.items()
+            }
+        else:
+            answer = form.data.get('award_or_cancel_decision')
+            if answer == 'back':
+                return redirect(url_for('.buyer_dos_requirements'))
```
idavidmcdonald: No flash message?

app/templates/buyers/cancel_brief.html
```diff
@@ -34,7 +34,7 @@
           {%
```
idavidmcdonald: If you visit this route from the cancel_or_award flow you will end up at the url ending `cancel-award`.

If you don't pick an answer, your form will be invalid when you submit it. It will be submitted to the `url_for('.cancel_brief')` which will then re render the at the url ending `cancel` with different title and hint text to the page you started at?

Is this correct or have I made a mistake/missed something?

tests/main/views/test_buyers.py
```diff
@@ -3719,3 +3778,116 @@ def test_buyer_account_overview_page_renders(self):
...
+
+    def test_back_redirects_to_buyer_dos_requiremnents_list(self):
+        self.login_as_buyer()
+        res = self.client.post(self.url.format(brief_id=self.brief['id']), data={'award_or_cancel_decision': 'back'})
+
+        redirect_text = html.fromstring(res.get_data(as_text=True)).text_content().strip()
+        expected_url = '/buyers/requirements/digital-outcomes-and-specialists'
+
+        assert res.status_code == 302
+        assert expected_url in redirect_text
```
katstevens: I think this redirect location assertion got missed off (the others look 👍 )

app/templates/buyers/cancel_brief.html
benvand: Fixed here:
https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/31/files#diff-98d4a0916d4357714af043c5b6347367R32

Uses `request.endpoint` like the view

benvand: Also tests:
https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/31/files#diff-9a0721b5c761c2fccb366192a3212493R3634
https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/31/files#diff-9a0721b5c761c2fccb366192a3212493R3676

app/templates/buyers/award_or_cancel_brief.html
benvand: Done

PR #33, opened 2017-09-18
Updated content for MVP

karlchillmaid:
Updated content for MVP: added missing word `participants` to the DOS heading.

Ticket 729: https://trello.com/c/xAmyH5CE/729-content-check-and-update-buyer-account-page

Updated the breadcrumb for the DOS requirements page.

Ticket 731: https://trello.com/c/03kbaHWk/731-content-dos-overview-page

### DIFF
--- a/app/main/helpers/buyers_helpers.py
+++ b/app/main/helpers/buyers_helpers.py
@@ -93,7 +93,7 @@ def get_briefs_breadcrumbs(additional_breadcrumbs=[]):
         },
         {
             "link": url_for("buyers.buyer_dos_requirements"),
-            "label": "Your requirements"
+            "label": "Requirements"
         }
     ]
 
--- a/app/templates/buyers/index.html
+++ b/app/templates/buyers/index.html
@@ -36,7 +36,7 @@
       {% endif %}
     </p>
   
-    <h2 class="heading-xmedium">Digital outcomes, specialists and user research</h2>
+    <h2 class="heading-xmedium">Digital outcomes, specialists and user research participants</h2>
     <p>
       {% if user_briefs_total %}
         <a href="{{ url_for('buyers.buyer_dos_requirements') }}">View your requirements</a><br>  
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -260,7 +260,7 @@ class BaseApplicationTest(object):
         breadcrumbs_we_expect = [
             ('Digital Marketplace', '/'),
             ('Your account', '/buyers'),
-            ('Your requirements', '/buyers/requirements/digital-outcomes-and-specialists'),
+            ('Requirements', '/buyers/requirements/digital-outcomes-and-specialists'),
         ]
         if extra_breadcrumbs:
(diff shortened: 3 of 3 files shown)


app/main/helpers/buyers_helpers.py
```diff
@@ -93,7 +93,7 @@ def get_briefs_breadcrumbs(additional_breadcrumbs=[]):
         },
         {
             "link": url_for("buyers.buyer_dos_requirements"),
-            "label": "Your requirements"
+            "label": "Requirements"
```
idavidmcdonald: Do we elsewhere use a different breadcrumb to the title of the page?

I believe this is the current convention (although please double check), so just checking you wish to diverge from this?

idavidmcdonald: This change also breaks some unit tests

karlchillmaid: This is different to what we usually do, but the benefits are it keeps the breadcrumb shorter, avoids the repetition of the word 'your' in the breadcrumb and keeps the breadcrumb more consistent between the DOS and G-Cloud sections of the buyer account.
I've spoken to Tobi and Jane about this and they've okayed it.

karlchillmaid: Is this because the unit tests are checking the content or because changing the content has a knock-on effect on other parts of the app/service?

idavidmcdonald: Cool, sounds like it's all OK then.

Of note, I just spotted a place that we also don't match the title with a breadcrumb so it looks like you aren't the first.

idavidmcdonald: It is because the unit tests are asserting what the breadcrumbs are.

The corresponding unit tests for this content can be found here:https://github.com/alphagov/digitalmarketplace-briefs-frontend/blob/9ed455b4e891253e31e4f2583d5868f5b9e7c094/tests/helpers.py#L263

You should just be able to make the identical content change to that line in the test file.

Let me (or one of the devs on your team) if this is unclear of you need any help :)

karlchillmaid: This is something George picked up this morning. The tickets are going back into the backlog and we'll land these changes once the tests have been updated and passed too.
Thanks.

karlchillmaid: Updated line 263 of the test file as part of this PR.

DilwoarH: @idavidmcdonald - is this okay now?

galund:
@idavidmcdonald @DilwoarH for the record, chapter and verse on GDS git policy is here:

https://github.com/alphagov/styleguides/blob/master/git.md#branchingmerging-conventions

Specifically, "you should not rewrite commits that have been pushed" (there are some exceptions and I'm cool with those). In other words, once you're in code review, there is no obligation, and you are in general discouraged, from rebasing. I can't help that practice on this team was different for ages :-(

PR #39, opened 2017-09-25
Add flash banner when 'still shortlisting' on award_or_cancel

benvand:


benvand:
Yeah I was tempted and I guess it's a bit of a pedantic way to write unit tests but I like that if the test fails then it tells you exactly what is wrong and only tests the smallest unit possible.

PR #40, opened 2017-09-26
Flake8 extra

benvand:
`flake8-putty` is not yet compatible with putty > v3.0.0 so we're pinning the version for now.
https://github.com/jayvdb/flake8-putty/issues/14

This PR pins the version lower, updates the make command and applys some flake8 fixes.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -9,7 +9,7 @@ watchdog==0.8.3
 freezegun==0.3.4
 
 coverage==3.7.1
-flake8==3.4.1
+flake8==2.6.2
 flake8-putty==0.4.0
 pytest-cov==2.2.0
 python-coveralls==2.5.0
--- a/tests/main/views/test_buyers.py
+++ b/tests/main/views/test_buyers.py
@@ -3723,12 +3723,8 @@ class TestCancelBrief(BaseApplicationTest):
     def test_that_no_option_chosen_does_not_trigger_update(self):
         res = self.client.post(self.url.format(brief_id=123))
 
-        document = html.fromstring(res.get_data(as_text=True))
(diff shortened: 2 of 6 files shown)


requirements-dev.txt
```diff
@@ -9,8 +9,8 @@ watchdog==0.8.3
 freezegun==0.3.4
 
 coverage==3.7.1
-flake8==3.4.1
+flake8==2.6.2
```
TheDoubleK: Why are we going for an old version?  Shiny and new is good?

tests/main/views/test_buyers.py
```diff
@@ -3723,13 +3723,9 @@ def test_that_no_option_chosen_triggers_error(self):
     def test_that_no_option_chosen_does_not_trigger_update(self):
         res = self.client.post(self.url.format(brief_id=123))
 
-        document = html.fromstring(res.get_data(as_text=True))
-        validation_message = document.xpath('//span[@class="validation-message"]')[0].text_content()
```
TheDoubleK: Feels like there was some intent to check the content of the message, rather than deleting the unused thing it might be better to add an assertion here? (This can also count as a Flake8 fix I think - as it would no longer be an unused thing.)

requirements-dev.txt
benvand: Updated PR desc for this

tests/main/views/test_buyers.py
benvand: Ah looks like it was a copy paste error from the test above. That's checking the same content.

benvand:
Thanks for the review @TheDoubleK I've addressed the comments.

PR #45, opened 2017-10-19
Update intro text for digital specialists and digital outcomes

JaneEastwood:
These changes were made to make it clearer that specialists found through the digital marketplace can only work for buyers within the scope of the written requirements. It was requested by CCS to help users decide which framework to use.

### DIFF
--- a/app/templates/buyers/start_brief_info.html
+++ b/app/templates/buyers/start_brief_info.html
@@ -40,10 +40,24 @@
     <div class="column-two-thirds large-paragraph">
         <div class="marketplace-paragraph">
             <p>
+                 {% set digital_specialists_intro %}
+                        You can use the Digital Marketplace to find a specialist, such as a developer or a user researcher, for a specific piece of work.
+                    </p> <p>
+                        You need to write 'requirements' to describe the work and tell suppliers to provide a specialist. The specialist can’t work for you outside the scope of your written requirements.
+                {% endset %}
+     
+                 {% set digital_outcomes_intro %}
+                        You can use the Digital Marketplace to find a team to work on a digital outcome, such as a booking system or an accessibility audit. 
+                    </p> <p>
+                        You need to write 'requirements' to describe the situation or problem and ask suppliers to propose a solution and provide a team.
+                    </p> <p>
+                        The members of the team can’t work for you outside the scope of your written requirements.
+                {% endset %}
+                
                 {{
                     {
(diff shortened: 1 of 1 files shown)


app/templates/buyers/start_brief_info.html
```diff
@@ -40,10 +40,24 @@
     <div class="column-two-thirds large-paragraph">
         <div class="marketplace-paragraph">
             <p>
+                 {% set digital_specialists_intro %}
+                        You can use the Digital Marketplace to find a specialist, such as a developer or a user researcher, for a specific piece of work.
+                    </p> <p>
```
TheDoubleK: Having half sets of tags in the `set` blocks and the other half outside feels a bit weird. I think it would be nicer to define all three messages as complete blocks including the paragraph tags.

e.g. 
```
{% set digital_specialists_intro %}
<p> ... </p>
<p> ... </p>
{% endset %}

{% set digital_outcomes_intro %}
<p> ... </p>
<p> ... </p>
{% endset %}

{% set user-research-participants_intro %}
<p> ... </p>
{% endset %}
```

TheDoubleK: I suggest grabbing a nearby developer to help fix this :)

app/templates/buyers/start_brief_info.html
```diff
@@ -40,10 +40,24 @@
...
+                        You need to write 'requirements' to describe the work and tell suppliers to provide a specialist. The specialist can’t work for you outside the scope of your written requirements.
```
TheDoubleK: These should be "fancy" typographical quotes around 'requirements'.

app/templates/buyers/start_brief_info.html
JaneEastwood: Thanks Kev! Pea has helped me with this :-)

app/templates/buyers/start_brief_info.html
JaneEastwood: Hi Kev
The style is for single quotes when we define something - see, for example, the first line on this page: https://www.gov.uk/tax-sell-home

TheDoubleK: Yep single quotes - but they are curly, not straight.

JaneEastwood: Thanks Kev. I've updated now those, if you wouldn't mind checking it again...

PR #46, opened 2017-10-19
Dh 127 content update fix all content issues

DilwoarH:


karlchillmaid:
Hi @benvand 
After some discussion we've decided to make breadcrumbs match page titles – at least until we can do a bigger piece of work on how we use breadcrumbs on the Digital Marketplace, when (if ever) we can shorten/simplify them, and whether what we do is consistent with the latest GOV.UK thinking.

benvand:
👍

PR #48, opened 2017-10-23
Buyers can withdraw briefs

katstevens:
Trello ticket: https://trello.com/c/7kq7pSTr/83-allow-buyers-withdraw-open-opportunities

New link:
withdraw1
Confirm banner:
withdraw2
Success banner:
withdraw3

- Buyers can now withdraw live requirements from the Brief summary page, in the same way that they would delete a draft requirement. The view and tests are based almost entirely on what's currently there for the delete flow - there may well be some gaps.
- Pulls in changes from the API client (merged) and the Frontend Toolkit (now merged)
- Fixes occurrences of `requirement` that were not pluralised (now confirmed with Constance).

Should now be ready for merging, although be aware that some functional tests will break with the copy changes.

### DIFF
--- a/app/main/forms/cancel.py
+++ b/app/main/forms/cancel.py
@@ -3,10 +3,10 @@ from wtforms import RadioField, validators
 
 
 class CancelBriefForm(Form):
-    """Form for the buyer to tell us why they are cancelling the requirement
+    """Form for the buyer to tell us why they are cancelling the requirements
     """
     choices = [
-        ('cancel', 'The requirement has been cancelled'),
+        ('cancel', 'Your requirements have been cancelled'),
         ('unsuccessful', 'There were no suitable suppliers'),
     ]
     cancel_reason = RadioField(
--- a/app/main/helpers/buyers_helpers.py
+++ b/app/main/helpers/buyers_helpers.py
@@ -17,12 +17,13 @@ def get_framework_and_lot(framework_slug, lot_slug, data_api_client, allowed_sta
     return framework, lot
 
 
-def is_brief_correct(brief, framework_slug, lot_slug, current_user_id, allow_withdrawn=False):
+def is_brief_correct(brief, framework_slug, lot_slug, current_user_id, allow_withdrawn=False, allowed_statuses=None):
     return (
         brief['frameworkSlug'] == framework_slug
         and brief['lotSlug'] == lot_slug
         and is_brief_associated_with_user(brief, current_user_id)
         and (True if allow_withdrawn else not brief_is_withdrawn(brief))
+        and (brief['status'] in allowed_statuses if allowed_statuses else True)
     )
 
 
--- a/app/main/views/buyers.py
+++ b/app/main/views/buyers.py
@@ -206,6 +206,7 @@ def view_brief_overview(framework_slug, lot_slug, brief_id):
     content = content_loader.get_manifest(brief['frameworkSlug'], 'edit_brief').filter({'lot': brief['lotSlug']})
     sections = content.summary(brief)
     delete_requested = True if request.args.get('delete_requested') else False
+    withdraw_requested = True if request.args.get('withdraw_requested') else False
 
     content_loader.load_messages(brief['frameworkSlug'], ['urls'])
     call_off_contract_url = content_loader.get_message(brief['frameworkSlug'], 'urls', 'call_off_contract_url')
@@ -235,6 +236,7 @@ def view_brief_overview(framework_slug, lot_slug, brief_id):
         completed_sections=completed_sections,
         step_sections=[section.step for section in sections if hasattr(section, 'step')],
         delete_requested=delete_requested,
+        withdraw_requested=withdraw_requested,
         call_off_contract_url=call_off_contract_url,
         framework_agreement_url=framework_agreement_url,
         awarded_brief_response_supplier_name=awarded_brief_response_supplier_name,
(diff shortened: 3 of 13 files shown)


app/main/views/buyers.py
```diff
@@ -1028,6 +1030,29 @@ def delete_a_brief(framework_slug, lot_slug, brief_id):
...
+    get_framework_and_lot(
+        framework_slug,
+        lot_slug,
+        data_api_client,
+        allowed_statuses=['live', 'expired'],
+        must_allow_brief=True
+    )
+    brief = data_api_client.get_brief(brief_id)["briefs"]
+
+    if not is_brief_correct(brief, framework_slug, lot_slug, current_user.id) or brief['status'] != 'live':
```
samuelhwilliams: Would it be good to add `allowed_statuses` to this method so that it can be consistently called like with the `get_framework_and_lot` method? Seems like it might be neater than a separate check.

app/templates/buyers/brief_overview.html
```diff
@@ -20,12 +20,37 @@
...
+              message = "Are you sure you want to delete these requirements?",
+              type = "destructive",
+              action = '<button type="submit" class="button-destructive banner-action">Yes, delete</button>'|safe
+            %}
+              {% include "toolkit/notification-banner.html" %}
+            {% endwith %}
+          </form>
+        </div>
+      {% endif %}
+      {% if withdraw_requested %}
```
samuelhwilliams: We should look at centralising these patterns in the frontend-toolkit given that flash messages are now generally handled there.

https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/templates/flash_messages.html

That doesn't currently support flashes with actions, but we should add it if we need it so that the work doesn't have to be done again when another app (inevitably) wants to do the same thing.

tests/main/views/test_buyers.py
```diff
@@ -647,9 +653,10 @@ def test_wrong_lot_get_view_brief_responses(self):
     def test_wrong_lot_publish_brief(self):
         self._load_page("/digital-outcomes/1234/publish", 404)
 
-    def test_wrong_lot_post_delete_a_brief(self):
-        data = {"delete_confirmed": True}
-        self._load_page("/digital-outcomes/1234/delete", 404, method='post', data=data)
+    @pytest.mark.parametrize('action', ['delete', 'withdraw'])
+    def test_wrong_lot_post_delete_or_withdraw_a_brief(self, action):
+        data = {"{}_confirmed".format(action): True}
+        self._load_page("/digital-outcomes/1234/{}".format(action), 404, method='post', data=data)
```
samuelhwilliams: This method reads badly in diffs. Doesn't describe what it's doing at all. It's possible to _guess_ that it's asserting it returns 404, but eh. Not really your fault of course... but it'd be nice to make this more transparent at some point.

And huh, having looked at the method, it does _a lot_ for it just being described as 'load page'.

katstevens: I am more than happy to refactor this!

app/main/forms/cancel.py
```diff
@@ -3,10 +3,10 @@
 
 
 class CancelBriefForm(Form):
-    """Form for the buyer to tell us why they are cancelling the requirement
+    """Form for the buyer to tell us why they are cancelling the requirements
     """
     choices = [
-        ('cancel', 'The requirement has been cancelled'),
+        ('cancel', 'The requirements have been cancelled'),
```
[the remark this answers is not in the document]
katstevens: This is one of the content things I need to check with Constance 😸

app/templates/buyers/brief_overview.html
katstevens: I was wondering whether that should be in scope of this story. I made the (reluctant) choice to pass through `"<some html>"|safe` to the toolkit for now, as it's difficult to know where this pattern is used elsewhere and what would break with the change. Also, how much to include in the new pattern - just the action button text? CSRF tokens/POST URL? The whole form div?

samuelhwilliams: It's hard to justify pulling it into a story because it does expand the scope a little bit, but I think we need to try to take positive actions like this wherever possible.

Things like CSRF should be centralised as much as possible IMO; the fewer places you need to implement anything security-focused, the better. That said, I don't immediately know how we'd design the interface for the action flash. It's probably not a small thing. :)

app/main/views/buyers.py
katstevens: Agreed - now done (for the other views that have Brief status checks as well).

app/main/views/buyers.py
```diff
@@ -1028,6 +1030,29 @@ def delete_a_brief(framework_slug, lot_slug, brief_id):
     return redirect(url_for('.buyer_dashboard'))
 
 
+@main.route('/frameworks/<framework_slug>/requirements/<lot_slug>/<brief_id>/withdraw', methods=['POST'])
+def withdraw_a_brief(framework_slug, lot_slug, brief_id):
+    get_framework_and_lot(
```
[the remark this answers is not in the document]
katstevens: We do use the returned framework and lot in a few of the views. I agree that hiding the 404s away in a helper are not great, but it's probably not one to untangle right now. Happy to add a tech debt ticket for it.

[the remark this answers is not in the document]
samuelhwilliams: 👍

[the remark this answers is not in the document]
katstevens: Now added to the tech debt board for triaging https://trello.com/c/KmpcxkaF/226-test-refactoring-helper-methods-in-briefs-fe-are-confusingly-named

app/main/forms/cancel.py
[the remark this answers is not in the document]
katstevens: This has now been changed to `Your requirements have been cancelled` (approved by Constance).

tests/main/views/test_buyers.py
katstevens: On closer inspection, this test class seems mostly redundant now. `_load_page` looks complex, but all that's really happening is a lot of setup (framework/content loader etc)  and a status code assertion for different combinations of lots and frameworks. The individual view classes check these pretty thoroughly. I'll tidy it up as best I can with the aim of removing the class entirely.

katstevens: Done. The diff isn't too bad, in the end. Hopefully the commit message is descriptive enough.

app/templates/buyers/brief_overview.html
katstevens: We do have a tech debt ticket (two, in fact) for this already https://trello.com/c/ZUmDyovj/23-flash-messages-dry-etc. I've added some details to that ticket that I mentioned above (CSRF etc).

app/main/views/buyers.py
samuelhwilliams: Nice! 👍

app/main/helpers/buyers_helpers.py
```diff
@@ -17,12 +17,13 @@ def get_framework_and_lot(framework_slug, lot_slug, data_api_client, allowed_sta
     return framework, lot
 
 
-def is_brief_correct(brief, framework_slug, lot_slug, current_user_id, allow_withdrawn=False):
+def is_brief_correct(brief, framework_slug, lot_slug, current_user_id, allow_withdrawn=False, allowed_statuses=None):
```
samuelhwilliams: Nice one. 👍

PR #51, opened 2017-10-30
Brief summary page hides banner for non-live withdrawals, non-draft deletes

katstevens:
Following QA of the withdraw ticket: https://trello.com/c/7kq7pSTr/83-allow-buyers-withdraw-open-opportunities

- Strengthens test assertions for when the delete/withdraw banners are shown (and when they should be hidden)
- Only draft briefs should show the 'Are you sure you want to delete these requirements' banner
- Only live briefs should show the 'Are you sure you want to withdraw these requirements' banner

### DIFF
--- a/app/main/views/buyers.py
+++ b/app/main/views/buyers.py
@@ -205,8 +205,8 @@ def view_brief_overview(framework_slug, lot_slug, brief_id):
 
     content = content_loader.get_manifest(brief['frameworkSlug'], 'edit_brief').filter({'lot': brief['lotSlug']})
     sections = content.summary(brief)
-    delete_requested = True if request.args.get('delete_requested') else False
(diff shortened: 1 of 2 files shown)


TheDoubleK:
Looks good. My only wonder is if 404 is the right error code... Feels a bit weird for the presence of a GET param to flip a working URL to a "not found". But 400 doesn't feel quite right either (although we misuse 400 a lot in the API). 

The other option could be to render the page as usual and ignore the param if brief not in the right status to be deleted/withdrawn? But 404 is probably fine.

TheDoubleK:
It's just that any other random param hacked into the URL would be ignored and page shown as usual. Having one "special" param that will 404 the page feels odd.

katstevens:
Now updated as discussed 👍

TheDoubleK:
Definitely better I reckon :+1:

PR #58, opened 2017-12-05
Delete css

MuriloDalRi:
https://trello.com/c/kGrDsUbw/80-remove-old-scss-briefs-front-end

* Application.css reduced from 238KB to 128KB before gzip
* Some of the imports from the toolkit are not currently being used so we have removed them so the application.js only imports things it actually uses.
* We haven't touched or really looked at _footer.scss as there are already plans to consolidate this.


app/assets/scss/_search-page.scss
```diff
@@ -1,92 +0,0 @@
-@import "search/_search-summary.scss";
-
-.search-summary {
```
samuelhwilliams: We also have `p.instruction-list-item-top.search-summary` in `instruction-list.scss`, which can probably go.

samuelhwilliams: Also it's interesting, because this class is used by `templates/toolkit/search-summary.html`, which we aren't removing (and is brought in by the toolkit). So now we have a template that our toolkit brings in but doesn't have any styling with it. Not sure what we think about this.

app/assets/scss/_publication.scss
```diff
@@ -1,383 +1,4 @@
...
-    margin-left: 0;
-  }
-
-}
-
-.floated-children {
-  overflow: hidden;
-}
 
 .terms-and-conditions {
```
idavidmcdonald: So I think we have the same issue with the cookies template for the terms and conditions template - we no longer need it as it is provided by the buyer FE.

Therefore I think we should delete the terms and conditions template as part of this PR. 

With that being the case, there is only one other use of the `terms-and-conditions` class which is on the link to the terms and conditions page in the footer template. Looking at this particular styling, it does not apply to it (I think we dealt with the same situation on one of the other apps). Therefore we should also be able to remove the whole publication.scss file.

app/assets/scss/application.scss
```diff
@@ -23,30 +23,21 @@ $path: "/static/images/";
 
 // Digital Marketplace Front-end toolkit styles
 @import "toolkit/_breadcrumb.scss";
-@import "toolkit/_browse-list.scss";
 @import "toolkit/_buttons.scss";
-@import "toolkit/_contact-details.scss";
-@import "toolkit/_document.scss";
```
idavidmcdonald: We are using this for the brief-responses.html template. This pulls in the toolkit documents template (https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/templates/documents.html) which requires the scss file for it.

On another note, github search sucks. If you search for `document` in the briefs repo you get nothing. If you search `documents` then you do find it...

app/assets/scss/_search-page.scss
idavidmcdonald: Yeah I'm not 100% sure on it either. I assume the template is being brought in by gulp which is copying all the templates?

It feels like there are three options
- change gulp to only bring in templates we need whilst also only bringing in the related scss
- only bring in scss we need but bring in all the templates
- bring in everything from both

All of them have pros and cons so I'm not entirely sure which I think is best yet. Any ideas?

samuelhwilliams: It feels pretty weird to have half of a component.

samuelhwilliams:
I've run this through Chrome's CSS coverage dev tool which reports we've still got about 45% unused CSS in our application.js after this, but it's pretty piecemeal and there aren't any obvious candidates for things that can be removed (i.e. some specific styles from one block aren't used, or some blocks from a component aren't used), so I think we've probably got all of the obvious stuff with this PR.

app/assets/scss/_publication.scss
MuriloDalRi: Done

app/assets/scss/application.scss
MuriloDalRi: 👍  I put it back

app/assets/scss/_search-page.scss
MuriloDalRi: I'm not sure what to do with this one. Should I put ```.search-summary back``` ?

samuelhwilliams: Not sure. I think maybe we need to have a chat around it. It's probably clearer for us doing this work if we remove it, as that's the principle we've applied everywhere else (ie just remove everything that isn't directly used by the app). We'll need to really clearly document and communicate our decisions around this (and the wider piece of work) though.

idavidmcdonald: Good idea. Let's talk about this issue tomorrow in person tomorrow. We can also talk about how we are making sure our decisions and new principles are documented and communicated.

app/assets/scss/application.scss
```diff
@@ -23,60 +23,33 @@ $path: "/static/images/";
...
-@import "toolkit/_temporary-message.scss";
 @import "toolkit/forms/_dates.scss";
 @import "toolkit/forms/_hint.scss";
-@import "toolkit/forms/_keyword-search.scss";
-@import "toolkit/forms/_option-select.scss";
 @import "toolkit/forms/_questions.scss";
 @import "toolkit/forms/_selection-buttons.scss";
 @import "toolkit/forms/_summary.scss";
 @import "toolkit/forms/_textboxes.scss";
 @import "toolkit/forms/_validation.scss";
```
idavidmcdonald: Near the top of the file we are importing `forms/word-counter` and `forms/list-entry` both from the toolkit. I think they could both be moved into this section with the rest of the toolkit imports (I can't think of a reason why the order would matter but worth just double checking).

MuriloDalRi: Are they being imported from the same place? Why don't they have ```toolkit/``` in the beggining?

app/assets/scss/_search-page.scss
MuriloDalRi: @samuelhwilliams can I remove this file then?

app/assets/scss/application.scss
idavidmcdonald: Maybe they are being imported from the other location (Sam mentioned there were two locations we were pulling in the toolkit, one of those being bower). I would assume it's safe to import them using the way we are doing the rest of our toolkit imports. However you can also spin the apps up and find a question with a word limit to check the word limit styling is working successfully.

idavidmcdonald:
If you could update the application.scss to look as close to our buyer FE and user FE versions as you possibly can too please for consistency.

app/assets/scss/application.scss
MuriloDalRi: Cool, I've changed that.

MuriloDalRi:
I think this is ready to go now

app/assets/scss/application.scss
idavidmcdonald: Nice. Also I did some reading (http://vanseodesign.com/css/sass-the-import-directive/) and confirmed that you can miss off the extension and the preceding underscore and sass still finds the file OK (not that it explains the likely importing from a different location stuff)

app/assets/scss/application.scss
```diff
@@ -1,82 +1,55 @@
...
-@import "toolkit/search/_search-result.scss";
+@import "toolkit/forms/list-entry";
```
idavidmcdonald: Can we do these consistently with the other tookit imports (adding a preceding underscore and the file extension). We could also order alphabetically too

samuelhwilliams: 👍 to consistency and alphabetisation.

app/assets/scss/_search-page.scss
idavidmcdonald: Did we reach a consensus on this @samuelhwilliams @MuriloDalRi or shall we speak about it in person to make sure we are all on the same page? BTW this was the thing I was trying to bring up yesterday morning but forgot the details

app/assets/scss/application.scss
MuriloDalRi: 👍

app/assets/scss/_search-page.scss
idavidmcdonald: Decided to not import the scss files we don't need regardless of if the related html templates are bulk imported by gulp.

app/assets/scss/application.scss
TheDoubleK: I have broken things alphabetising them in the past - import order can matter for scss. (I have no idea of its a potential issue here though.)

idavidmcdonald: Hopefully we will pick up if something has broken. I think in the long run we should be able to structure our toolkit and scss files such that alphabetical ordering should not cause any problems (our files should not have unexpected knock on effects for other components). In the meantime if we spot anything explicitly depending on non alphabetical order let's add it as an inline comment

PR #68, opened 2018-01-11
Update frameworks and safely load manifests

samuelhwilliams:
## Summary
Update frameworks version to 10.2.0 to bring in DOS3. Also catch errors (and log) when loading a manifest fails, so that the app can still come up.

## Ticket
https://trello.com/c/tnBdllTV/156-briefs-frontend-has-crashed-on-preview-wont-come-back-up

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -4,6 +4,7 @@ from flask_wtf.csrf import CsrfProtect
 
 import dmapiclient
 from dmutils import init_app, flask_featureflags
+from dmcontent.utils import try_load_manifest
 from dmcontent.content_loader import ContentLoader
 from dmutils.user import User
 
@@ -35,11 +36,12 @@ def create_app(config_name):
         if not framework_data['slug'] in application.config.get('DM_FRAMEWORK_CONTENT_MAP', {}):
             if framework_data['framework'] == 'g-cloud':
                 if framework_data['status'] != 'expired':
-                    content_loader.load_manifest(framework_data['slug'], 'services', 'search_filters')
-                # we need to be able to display old services, even on expired frameworks
-                content_loader.load_manifest(framework_data['slug'], 'services', 'display_service')
+                    try_load_manifest(content_loader, application, framework_data, 'services',
+                                      'services_search_filters')
+
+                try_load_manifest(content_loader, application, framework_data, 'services', 'display_service')
             elif framework_data['framework'] == 'digital-outcomes-and-specialists':
-                content_loader.load_manifest(framework_data['slug'], 'briefs', 'display_brief')
+                try_load_manifest(content_loader, application, framework_data, 'briefs', 'display_brief')
 
     from .create_buyer.views.create_buyer import create_buyer as create_buyer_blueprint
     from .main import dos as dos_blueprint
--- a/bower.json
+++ b/bower.json
@@ -7,6 +7,6 @@
     "jquery-details": "https://github.com/mathiasbynens/jquery-details/archive/v0.1.0.tar.gz",
     "digitalmarketplace-frontend-toolkit": "https://github.com/alphagov/digitalmarketplace-frontend-toolkit.git#26.3.0",
     "govuk_template": "https://github.com/alphagov/govuk_template/releases/download/v0.19.2/jinja_govuk_template-0.19.2.tgz",
-    "digitalmarketplace-frameworks": "https://github.com/alphagov/digitalmarketplace-frameworks.git#9.1.0"
+    "digitalmarketplace-frameworks": "https://github.com/alphagov/digitalmarketplace-frameworks.git#10.2.0"
   }
 }
--- a/requirements-app.txt
+++ b/requirements-app.txt
@@ -10,5 +10,5 @@ Flask-WTF==0.12
 git+https://github.com/alphagov/odfpy.git@ee4482a#egg=odfpy==1.3.6dev
 
 git+https://github.com/alphagov/digitalmarketplace-utils.git@31.0.0#egg=digitalmarketplace-utils==31.0.0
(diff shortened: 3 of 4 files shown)


samuelhwilliams:
```
/Users/samuelwilliams/git/digitalmarketplace-briefs-frontend/venv/bin/flake8 .
./app/__init__.py:22:1: C901 'create_app' is too complex (15)
make: *** [test-flake8] Error 1
```

:man_facepalming:

app/__init__.py
```diff
@@ -31,15 +32,26 @@ def create_app(config_name):
         login_manager=login_manager,
     )
 
+    def try_load_manifest(framework_data, question_set, manifest):
+        try:
+            content_loader.load_manifest(framework_data['slug'], question_set, manifest)
+
+        except ContentNotFoundError:
+            # Not all frameworks have this, so no need to panic (e.g. G-Cloud 4, G-Cloud 5, G-Cloud-6)
```
idavidmcdonald: This comment doesn't make much sense when this logic is abstracted to it's own function. If anything, the comment should go with where we decide to call `try_load_manifest`, defining why we think it's ok to use the function.

app/__init__.py
```diff
@@ -18,6 +19,17 @@
 content_loader = ContentLoader('app/content')
 
 
+def _try_load_manifest(application, data, question_set, manifest):
+    try:
+        content_loader.load_manifest(data['slug'], question_set, manifest)
+
+    except ContentNotFoundError:
+        # Not all frameworks have this, so no need to panic (e.g. G-Cloud 4, G-Cloud 5, G-Cloud-6)
```
idavidmcdonald: This comment doesn't make any sense when this functionality is abstracted into it's own function. If anything, this comment should be made when we decide to use `_try_load_manifest` as you need the context (it doesn't apply for every time we run the helper func).

app/__init__.py
```diff
@@ -35,11 +47,12 @@ def create_app(config_name):
         if not framework_data['slug'] in application.config.get('DM_FRAMEWORK_CONTENT_MAP', {}):
             if framework_data['framework'] == 'g-cloud':
                 if framework_data['status'] != 'expired':
-                    content_loader.load_manifest(framework_data['slug'], 'services', 'search_filters')
```
idavidmcdonald: I'm not sure if we should go the approach of hiding errors by default. Instead can we build further on the statuses we expect a manifest to exist for? e.g. not for open or expired frameworks (or whatever it is). It feels better to be explicit about what state we want need our manifests to be in for an app to start rather than assume it is set up in that way). Saying that, I had assumed that unit tests would have picked up on this rather than us finding out in the way we did which wasn't ideal. 

If we did decide to use the try_load then we could also remove the framework status checks as they are essentially redundant.

Feel free to argue the counter these points, my opinions aren't too strongly held.

idavidmcdonald: Also just to double check, for the occasions we are using try load instead of just loading, must we do so every time (e.g. can be more explicit sometimes and be less others based on framework type/state)

app/__init__.py
```diff
@@ -18,6 +19,17 @@
 content_loader = ContentLoader('app/content')
 
 
+def _try_load_manifest(application, data, question_set, manifest):
```
TheDoubleK: Of course, this is the way to do it :+1:

app/__init__.py
TheDoubleK: I don't think the comment quite works here though, because it could be anything it's trying to load, even something that all frameworks should have...

bower.json
```diff
@@ -7,6 +7,6 @@
     "jquery-details": "https://github.com/mathiasbynens/jquery-details/archive/v0.1.0.tar.gz",
     "digitalmarketplace-frontend-toolkit": "https://github.com/alphagov/digitalmarketplace-frontend-toolkit.git#26.3.0",
     "govuk_template": "https://github.com/alphagov/govuk_template/releases/download/v0.19.2/jinja_govuk_template-0.19.2.tgz",
-    "digitalmarketplace-frameworks": "https://github.com/alphagov/digitalmarketplace-frameworks.git#9.1.0"
+    "digitalmarketplace-frameworks": "https://github.com/alphagov/digitalmarketplace-frameworks.git#10.2.0"
```
TheDoubleK: Belt and braces!  Why not...

samuelhwilliams:
@benvand made a good point in that the try_load_manifest thing should be in the content-loader, so I'm going to move it there. It means an extra PR and I need to write a test or two, but that's probably the better place to put it. Bear with.

app/__init__.py
TheDoubleK: So I don't think it's possible to define by framework status what we'd expect to be there.  I guess the thing we're avoiding by doing the `try` for everything is having to update content in all of the frontend repos before you can add a new framework even only on Preview.

Maybe that's being  lazy and we should have the new framework everywhere, I don't know.  But feels to me like for applications to open we should only need buyer and supplier apps to have the content, as they are the only apps that do anything to a framework before applications are closed.

samuelhwilliams:
https://github.com/alphagov/digitalmarketplace-content-loader/pull/54

app/__init__.py
samuelhwilliams: I don't think building out the statuses bit here is necessarily a good idea (and if we were to go that route, that's something that would need to start to be centralised, probably). Try/load takes us a step towards consistency (Eg with the admin-frontend) and limits the damage the problem can cause in the future. If we want to take a closer look at this, it's probably something we should do as tech debt/scalability issue.

idavidmcdonald: Cool, sounds like everyone else is also behind that approach so I'll follow your lead 👍

PR #77, opened 2018-02-13
Revert back to eea/odfpy

samuelhwilliams:
Update utils to use eea/odfpy rather than our interim alphagov/odfpy
fork, which we created to allow us to patch a bug. Now that the bug has
been accepted into the main library as of release 1.3.6
(https://github.com/eea/odfpy/pull/72) we no longer need to support own
our fork.

Also adds a comment when registering the external blueprint from 
dm-utils to ensure it is registered last.


TheDoubleK:
Oh, you've already done the changelog thing.  Fine then :p

samuelhwilliams:
@galund I guess, although it will only duplicate what's in the comment... :man_shrugging:

PR #81, opened 2018-02-21
Use FE toolkit analytics files to set custom dimensions

brenetic:
The `digitalmarketplace-frontend-toolkit` should now be home to the more generic GA related stuff. `_events.js` holds all the app specific stuff required by Briefs Frontend.

https://trello.com/c/Jj4cz4u4

Merge after
- [x] https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/408

Paired with @katstevens


app/templates/_custom_meta.html
```diff
@@ -0,0 +1,3 @@
+{% block custom_meta %}
```
galund: I'm not sure I understand why this is in a `{% block %}` - if on some page we decide we want to add additional meta tags, we wouldn't want custom dimension taken out, would we? Maybe I'm missing the point here... also I wonder if this should be in the FE-toolkit?

katstevens: That's a good spot. (The pattern's copied from the Buyer FE where we set the custom meta on one particular page.) 

We should do something like this instead:
```
_custom_meta.html:

<meta name="ga_customDimension" data-id="10" data-value="{{ current_user.role }}">
{% block custom_meta %}{% endblock %}
```

katstevens:
OK, I think the following usage would work:

`_base_page.html`
```
{% block head %}
    {% with custom_dimensions = [
	{“data-id”: 10, “data-value”: “supplier”},
	{“data-id”: 11, “data-value”: “SME”},
    ] %}
        {% include toolkit/custom-meta.html %}
    {% endwith %}
{% endblock %}
```

And the `toolkit/custom_meta.html` as per https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/408/commits/4f7355b3768e58fef099f6900321cee57d7ad68f

app/templates/_base_page.html
```diff
@@ -1,7 +1,11 @@
 {% extends "govuk/govuk_template.html" %}
 
 {% block head %}
-  {% include "_custom_meta.html" %}
+  {% with custom_dimensions = [
+    {"data_id": 10, "data_value": current_user.role }
```
galund: Potentially the association of a particular dimension id (i.e. 10) with a particular field (i.e. user.role) should also be centralized. Whatever happens, we can't ever redefine that in another app.

Do all our apps put `current_user` into the template? If so, maybe this can literally be lifted into toolkit template rather than passed in.

katstevens: I think it's in the context for all apps, however after checking with Ash, we don't want to set this particular dimension if the user isn't logged in (i.e. where `current_user.role == None`). We could do a null check on `data-value` in the toolkit but it feels like that logic should be at app level. 

Also this means we can reuse the `custom-dimensions.html` template for other dimensions set on specific URLs (e.g. the saved search status on the Buyer FE), without having to worry about duplicating the user role dimension on the same page. 

I agree that having a centralised list of dimensions and their IDs is sensible in the long run, but it's probably out of scope for this story?

galund: You're very much relying here on the fact that the briefs app is logged-in-only. In general we might be redirecting people to any app (well, particularly in the case of buyers, they might be redirected to a page on the buyer app, e.g. G-Cloud pages which are available both logged-in and logged-out).

In which case you will anyway have to put an `if` into the template to do the check for the not-logged-in case -- and that would be the same in the other apps, at least if they needed to handle both logged in and not.

It definitely feels to me like it might be the right thing to do to centralize this from the off?

katstevens:
The Snyk warning went away when I pulled in `28.2.0` instead of `v28.2.0` - I don't understand why though 😿

spec/javascripts/unit/AnalyticsSpec.js
```diff
@@ -215,125 +215,4 @@ describe("GOVUK.Analytics", function () {
...
-    });
-
-    it('should not send the lot as a custom dimension for any non DOS lot', function() {
-      spyOn(GOVUK.GDM.analytics.location, "pathname")
-        .and
-        .returnValue('/not-digital-outcomes-and-specialists/opportunities/100');
-      window.GOVUK.GDM.analytics.pageViews.init();
-      expect(window.ga.calls.all().map(function (i) {return i.args})).not.toContain(['set', 'dimension26', 'test-lot']);
-    });
-  });
```
risicle: Though this is more a test of the code in the toolkit now, is it worth still keeping this in here as an integration test of sorts - to ensure the toolkit code is being used right?

If it would mean lots of duplicated code remaining in the various frontend repos, the common body of the test could be moved to the toolkit...

katstevens: In theory yes, however these specific tests aren't actually testing the Briefs FE! It's all public DOS pages and searches, copied over boilerplate from the Buyer FE.

risicle: God damn that's weird

PR #82, opened 2018-02-23
Move flash messages out of template

galund:
Flash messages moving out of templates and into Python code. I've taken the opportunity to make them have nice curly quotes, too.

Also moving some shared test login code into dm-test-utils.

https://trello.com/c/lQIP4rf8/300-flash-messages-briefs-app

### DIFF
--- a/app/templates/buyers/dashboard.html
+++ b/app/templates/buyers/dashboard.html
@@ -30,15 +30,7 @@
                             <div class="banner-success-without-action">
                         {% endif %}
                                 <p class="banner-message">
-                        {% if 'requirements_deleted' in message %}
-                                Your requirements &lsquo;{{ message.get('requirements_deleted') }}&rsquo; were deleted
-                        {% elif 'requirements_withdrawn' in message %}
-                                You've withdrawn your requirements for &lsquo;{{ message.get('requirements_withdrawn') }}&rsquo;.
-                        {% elif 'updated-brief' in message %}
-                                You've updated '{{ message.get('updated-brief') }}'
-                        {% else %}
-                            {{ message }}
-                        {% endif %}
+                                    {{ message }}
                                 </p>
                             </div>
                     {% endfor %}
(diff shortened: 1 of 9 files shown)


app/templates/buyers/dashboard.html
```diff
@@ -30,15 +30,7 @@
...
-                        {% if 'requirements_deleted' in message %}
-                                Your requirements &lsquo;{{ message.get('requirements_deleted') }}&rsquo; were deleted
-                        {% elif 'requirements_withdrawn' in message %}
-                                You've withdrawn your requirements for &lsquo;{{ message.get('requirements_withdrawn') }}&rsquo;.
-                        {% elif 'updated-brief' in message %}
-                                You've updated '{{ message.get('updated-brief') }}'
-                        {% else %}
-                            {{ message }}
-                        {% endif %}
+                                    {{ message }}
```
idavidmcdonald: Is it worth taking this time to refactor the few lines around this to follow the structure of `brief_overview.html` template so it's also using the toolkit macro rather than raw HTML?

At a glance it seems possible and I assume wouldn't change any behaviour but worth double checking

galund: Yeah I had a go and it was going to cause some layout issues (too much vertical whitespace seems to be an easy thing to end up with).

So I decided it made more sense to do all of that in one go once the template gets centralized.

galund: Captured here: https://trello.com/c/8Ix9Q5vn/371-flash-messages-centralize-the-template-all-apps

tests/main/views/test_buyers.py
```diff
@@ -3905,18 +3893,7 @@ def test_redirect_on_successful_status_change(self, status):
...
-
-        with self.client.session_transaction() as session:
-            session['_flashes'] = [
-                ('message', {'updated-brief': "My Amazing Brief"})
-            ]
-        expected_url = '/buyers/frameworks/digital-outcomes-and-specialists-2/requirements/digital-outcomes/123'
-        res = self.client.get(expected_url)
-        flash_div = html.fromstring(res.get_data(as_text=True)).xpath('//div[@class="banner-success-without-action"]')
-        assert flash_div[0].text_content().strip() == "You've updated 'My Amazing Brief'"
+        self.assert_flashes("You’ve updated ‘I need a thing to do a thing’", "message")
```
katstevens: 👍 for getting rid of the `session['_flashes']` bits in the tests!

PR #93, opened 2018-03-29
Briefs FE test refactor

katstevens:
The aim here is to have the Briefs tests be consistent with each other. This means refactoring the following:

- Patching the API client consistently, at class setup/teardown level (eventually these could go in the `BaseApplicationTest` setup/teardown?)
- Not using `with self.app_context` where it's not needed (at least, where the tests happily pass without it!)
- Having shared setup for parts of the test where possible. This should help us see what we could theoretically move to test utils or the base class, and what needs to be kept at app/module/test class/unit test level.
- We don't need to specify the test class name in `super()` in Py3.

This PR may look horrendous (it was quick to actually do in the end) but the commits are sanely ordered by class (or module, if it's a short one) so hopefully will be manageable. 

Not included this time: `test_download_responses.py` (this covers a class based view which patches things weirdly. This branch is big enough as it is so I decided to leave it for now).

If folks agree this test style is sensible, I will do the same thing across the other FE apps to get some consistency.


tests/create_buyers/views/test_create_buyers.py
```diff
@@ -126,13 +131,14 @@ def test_email_address_is_correctly_stored_in_session(self, data_api_client, DMN
...
             data={'email_address': 'valid@test.gov.uk'},
             follow_redirects=False
         )
         assert res.status_code == 302
-        data_api_client.create_audit_event.assert_called_with(audit_type=AuditTypes.invite_user,
-                                                              data={'invitedEmail': 'valid@test.gov.uk'})
+        self.data_api_client.create_audit_event.assert_called_with(
+            audit_type=AuditTypes.invite_user,
+            data={'invitedEmail': 'valid@test.gov.uk'}
+        )
```
benvand: +1 for keeping indentation levels at multiples of 4 💯

tests/main/views/test_buyers.py
```diff
@@ -85,15 +80,20 @@ def find_briefs_mock():
...
 
-@mock.patch('app.main.views.buyers.data_api_client', autospec=True)
 class TestBuyerDashboard(BaseApplicationTest):
 
     def setup_method(self, method):
-        super(TestBuyerDashboard, self).setup_method(method)
+        super().setup_method(method)
+        self.data_api_client_patch = mock.patch('app.main.views.buyers.data_api_client', autospec=True)
+        self.data_api_client = self.data_api_client_patch.start()
+        self.data_api_client.find_briefs.return_value = find_briefs_mock()
```
benvand: This is a nice way of doing a patch. It's a little less obvious but it's cleaner.

PR #99, opened 2018-04-20
Move Briefs FE flash templates into base template

DilwoarH:
Epic: https://trello.com/c/5pZdW4Cb/122-centralise-frontend-header-footer-templates

Currently the templates include `toolkit/flash_messages.html` in a couple of different places.

Moving it to the base template has previously caused weird whitespace issues - these should be easier to sort out now we've moved out some analytics logic from the templates.

Ideally the Briefs FE base template should be consistent with the other FE apps.

Ticket: https://trello.com/c/0q1xiDvF/47-move-briefs-fe-flash-templates-into-base-template

### DIFF
--- a/app/templates/_base_page.html
+++ b/app/templates/_base_page.html
@@ -25,6 +25,7 @@
   {% block breadcrumb %}{% endblock %}
   <div id="wrapper">
     <main id="content" role="main">
+      {% include "toolkit/flash_messages.html" %}
       {% block main_content %}
       {% endblock %}
     </main>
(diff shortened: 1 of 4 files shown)


galund:
So, yes this is the change I'm expecting to see, but I suppose I'm wary that I kinda expected _some_ layout problems to need fixing. Adding @katstevens to this PR just in case she knows where to look based on earlier experimentation with this stuff.

DilwoarH:
Hey @katstevens :)

PR #102, opened 2018-04-27
Move buyer app templates into frontend-toolkit

DilwoarH:
Epic: https://trello.com/c/5pZdW4Cb/122-centralise-frontend-header-footer-templates

* Buyer app templates such as `_base_page.html` move into the frontend-toolkit.
* Organise the frontend-toolkit the frontend toolkit so it's easy to distinguish between its existing set of 'components' and templates like this that we are just including or extending from other DM apps;
* Move any Javascript that's easy to move but we can mostly come back to refactoring of that in another ticket.

Ticket: https://trello.com/c/lzYdOQ4C/56-move-buyer-app-templates-into-frontend-toolkit


tests/main/views/test_notifications.py
```diff
@@ -16,7 +16,6 @@ def test_should_see_banner_if_not_subscribed(self):
         self.login_as_buyer(user_research_opted_in=False)
         res = self.client.get('/buyers')
         assert res.status_code == 200
-        assert 'Help us improve the Digital Marketplace' in res.get_data(as_text=True)
```
samuelhwilliams: I think you've removed this by accident. This is slightly weird as the exact same content is used for two separate things: a flash/header asking users to become research participants, and a page footer asking if there's anything wrong with the page.

This test is checking the user research participant banner, and the content hasn't changed, so this assertion should remain in.

DilwoarH: hey sam,

I've added it back in but with a check to specifically look for the banner-heading

PR #109, opened 2018-05-27
Refactor validation masthead use

samuelhwilliams:
## Summary
We have an updated version of the validation masthead for pages
displaying errors with form submissions that abstracts some more of the
responsibility from the individual templates and supports our main error
formats natively (content loader forms and wtforms). This means we can
rip out some of the boilerplate from our templates.

There's also a page here that didn't use the validation masthead from
the frontend-toolkit which has now been brought in line.

 ## Ticket
https://trello.com/c/RSYEM3FL/448

### DIFF
--- a/app/main/views/outcome.py
+++ b/app/main/views/outcome.py
@@ -14,6 +14,7 @@ from ..forms.cancel import CancelBriefForm
 from ..forms.award_or_cancel import AwardOrCancelBriefForm
 
 from dmapiclient import HTTPError
+from dmutils.forms import get_errors_from_wtform
 
 BRIEF_UPDATED_MESSAGE = "You’ve updated ‘{brief[title]}’"
 
@@ -55,10 +56,7 @@ def award_or_cancel_brief(framework_slug, lot_slug, brief_id):
         if request.method == "POST":
             form = AwardOrCancelBriefForm(brief, request.form)
             if not form.validate_on_submit():
-                errors = {
-                    key: {'question': form[key].label.text, 'input_name': key, 'message': form[key].errors[0]}
-                    for key, value in form.errors.items()
-                }
+                errors = get_errors_from_wtform(form)
             else:
                 answer = form.data.get('award_or_cancel_decision')
                 if answer == 'back':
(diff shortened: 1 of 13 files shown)


app/main/views/outcome.py
```diff
@@ -55,10 +56,7 @@ def award_or_cancel_brief(framework_slug, lot_slug, brief_id):
         if request.method == "POST":
             form = AwardOrCancelBriefForm(brief, request.form)
             if not form.validate_on_submit():
-                errors = {
-                    key: {'question': form[key].label.text, 'input_name': key, 'message': form[key].errors[0]}
-                    for key, value in form.errors.items()
-                }
+                errors = get_errors_from_wtform(form)
```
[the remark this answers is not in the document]
samuelhwilliams: 👍

app/main/views/outcome.py
```diff
@@ -48,36 +46,32 @@ def award_or_cancel_brief(framework_slug, lot_slug, brief_id):
...
+            # We should never get here as the form validates the answers against the available choices.
+            abort(500, "Unexpected answer to award or cancel brief")
+
+    errors = get_errors_from_wtform(form)
 
     return render_template(
         "buyers/award_or_cancel_brief.html",
         brief=brief,
-        form=form or AwardOrCancelBriefForm(brief),
+        form=form,
```
katstevens: Lovely. Could we do the same for `award_brief` and `cancel_brief`? 😸

samuelhwilliams: 😢

samuelhwilliams: Done 👍

app/main/views/outcome.py
```diff
@@ -47,39 +46,32 @@ def award_or_cancel_brief(framework_slug, lot_slug, brief_id):
...
-                        url_for('.award_brief', framework_slug=framework_slug, lot_slug=lot_slug, brief_id=brief_id)
-                    )
-                elif answer == 'no':
-                    return redirect(url_for(
-                        '.cancel_award_brief', framework_slug=framework_slug, lot_slug=lot_slug, brief_id=brief_id)
-                    )
-                else:
-                    # We should never get here as the form validates the answers against the available choices.
-                    abort(500, "Unexpected answer to award or cancel brief")
+    form = AwardOrCancelBriefForm(brief)
```
katstevens: Hmm. I think we still need the `if request.method == "POST": ... request.form` data in here somewhere (sorry, I think I missed that on my previous comment).

samuelhwilliams: I don't think we do?

`form.validate_on_submit` already checks that it's a POST method, and it also reads directly from `request.form` where appropriate to fill in the fields.

katstevens: That makes sense - I was wondering why the tests still passed 😹 If it means more code gets deleted then I'm all for it!

PR #126, opened 2018-08-16
Briefs FE error handling

katstevens:
Trello: https://trello.com/c/3iwy5gAy/115-pull-in-error-handling-to-fe-apps

Straightforward for this app:

- Pull in new utils/FE toolkit
- Remove old templates
- Remove old app level error handlers
- Adds content loader `QuestionNotFoundError` to be consistent with other FE apps
- Adds some tests. One 'unrelated' test (checking the X-Frame header) needed the status code changing due to a change in the test class setup (we now fully patch the Data API client so the view renders correctly). This brings it in line with the expected result for the same test on the other apps.


lfdebrux:
All looks good!

PR #129, opened 2018-08-24
remove use of FEATURE_FLAGS_NEW_SUPPLIER_FLOW, bump digitalmarketplace-utils to v43.0.0

risicle:
See https://trello.com/c/ZH9trgjH/419-snyk-alert-for-flask-010x-vulnerability

This one required a bit of fiddling to remove its use of feature flags. From the changeset comment:

> remove use of FEATURE_FLAGS_NEW_SUPPLIER_FLOW, replace with is_legacy_brief_response logic
>
> we want to retire the use of flask feature flags and this is one of the last uses of them. previously we were comparing the `publishedAt` date with a feature flag's `enabled_since` date to determine whether it's a "legacy brief" or not. instead use the logic of `is_legacy_brief_response`, adapted from brief-responses-frontend, to figure this out. this function works by inspecting a brief *response*, so doesn't work in cases where there are no responses. fortunately the only bit of template that want to know its legacy status make no sense if there aren't any brief responses, so that largely solves that. update tests to provide information needed by `is_legacy_brief_response` in mocks.


app/main/views/buyers.py
```diff
@@ -414,9 +419,10 @@ def view_brief_responses(framework_slug, lot_slug, brief_id):
 
     brief_responses = data_api_client.find_brief_responses(brief_id)['briefResponses']
 
-    brief_responses_require_evidence = (
-        datetime.strptime(current_app.config['FEATURE_FLAGS_NEW_SUPPLIER_FLOW'], "%Y-%m-%d")
-        <= datetime.strptime(brief['publishedAt'][0:10], "%Y-%m-%d")
+    brief_responses_required_evidence = (
+        None
+        if not brief_responses else
+        not is_legacy_brief_response(brief_responses[0], brief=brief)
```
katstevens: Cool 👍I think it's a better explanation of what's happening than the alternative I just thought of... 

... hardcoding a check for `brief['publishedAt'][0:10] > '2017-02-08'` - I mean, we wouldn't need separate checks for preview/staging as that data has long been wiped, but it feels a bit icky.

risicle: Yeah, mine still isn't an ideal way, in that `None` is falsey, but meh...

risicle:
Just waiting for the OK on this one now...

PR #135, opened 2018-09-25
Load supplier lot csvs from correct framework s3 bucket

Wynndow:
For [this trello ticket.](https://trello.com/c/6flmLBWC)

The csvs of suppliers by lot were being loaded from the original DOS
framework bucket location, regardless of the framework iteration. This is weird
and confusing and this fixes that.

The files have been copied to the DOS2 bucket, ready for this change. The new files for DOS3 will be generated and uploaded soon. Tomorrow.


Wynndow:
Well I am a superhero.

PR #156, opened 2019-01-11
Use API stubs

katstevens:
Pulls in the new API stubs from dmtestutils.

Notes for reviewers:
- I hope it's clear why both the `Stub.response()` and `Stub.single_api_response()` are needed - they reflect what happens in `Model.serialize()` versus the API view response.
- BriefStub().response() includes a `user_id=123` if users are required later on by the test (e.g. the stub is plonked into `{"briefs": self.brief}`. Normally the `users` key gets added at the `.single_result_response()` level, to reflect the API view response. 
- There's a lot of repetition in some of these tests, especially around framework/lot setup; I haven't refactored them in this PR as it's too big and I want the simple API stub changes to be visible.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -13,4 +13,4 @@ flake8-per-file-ignores==0.4.0
 pytest-cov==2.5.1
 python-coveralls==2.5.0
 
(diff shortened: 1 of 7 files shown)


risicle:
Well that *is* interesting.

PR #174, opened 2019-02-11
Update to weekly

jonodrew:


katstevens:
Ah - actually can we add the `# pyup: ignore` to itsdangerous here as well?

PR #230, opened 2019-10-30
Add govuk-frontend-jinja and digitalmarketplace-govuk-frontend

risicle:
https://trello.com/c/Jwpu5Zg2

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -14,6 +14,7 @@ const repoRoot = path.join(__dirname)
 const npmRoot = path.join(repoRoot, 'node_modules')
 const govukToolkitRoot = path.join(npmRoot, 'govuk_frontend_toolkit')
 const govukElementsRoot = path.join(npmRoot, 'govuk-elements-sass')
+const govukFrontendRoot = path.join(npmRoot, 'digitalmarketplace-govuk-frontend', 'govuk-frontend')
 const dmToolkitRoot = path.join(npmRoot, 'digitalmarketplace-frontend-toolkit', 'toolkit')
 const sspContentRoot = path.join(npmRoot, 'digitalmarketplace-frameworks')
 const assetsFolder = path.join(repoRoot, 'app', 'assets')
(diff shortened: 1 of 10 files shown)


gulpfile.js
```diff
@@ -236,6 +239,24 @@ gulp.task(
   )
 )
 
+gulp.task(
+  'copy:govuk_frontend_assets:fonts',
+  copyFactory(
+    'fonts from the GOVUK frontend assets',
+    govukFrontendFontsFolder,
+    assetsFolder + '/fonts'
```
lfdebrux: This should actually be `staticFolder`; see https://github.com/alphagov/digitalmarketplace-user-frontend/pull/176

risicle: ah ok

gulpfile.js
```diff
@@ -286,9 +309,9 @@ gulp.task('build:development', gulp.series(gulp.parallel('set_environment_to_dev
 gulp.task('build:production', gulp.series(gulp.parallel('set_environment_to_production', 'clean'), 'compile'))
 
 gulp.task('watch', gulp.series('build:development', function () {
-  const jsWatcher = gulp.watch([path.join(assetsFolder, '**', '*.js')], gulp.series('js'))
-  const cssWatcher = gulp.watch([path.join(assetsFolder, '**', '*.scss')], gulp.series('sass'))
-  const dmWatcher = gulp.watch([path.join(npmRoot, 'digitalmarketplace-frameworks', '**')], gulp.series('copy:frameworks'))
+  const jsWatcher = gulp.watch([assetsFolder + '/**/*.js'], gulp.series('js'))
+  const cssWatcher = gulp.watch([assetsFolder + '/**/*.scss'], gulp.series('sass'))
+  const dmWatcher = gulp.watch([npmRoot + '/digitalmarketplace-frameworks/**'], gulp.series('copy:frameworks'))
```
aliuk2012: Why have we stopped using `path.join`?

risicle: I must have copypasta'd it from an example that doesn't... and sure enough the user-frontend doesn't either in these lines.

aliuk2012: User-frontend should use path.join. I'll do a pr for that. could you revert these cahnges?

PR #232, opened 2019-11-01
switch most buttons to use digitalmarketplace-govuk-frontend

risicle:
https://trello.com/c/4vOITu1f

~@lfdebrux it didn't turn out to be easy to switch the in-summary-table buttons without breaking out into raw html. The summary table macros really don't seem to have any capacity for arbitrary content.~ This is not true - it's just the ability to do so doesn't really have any examples in the documentation - I'll have another go...

Pages all (FSVO "all" - I could have missed some cases in my checking) seem to work and look approximately right.

### DIFF
--- a/app/templates/buyers/award.html
+++ b/app/templates/buyers/award.html
@@ -44,14 +44,10 @@
                     {% endwith %}
 
                     {% block save_button %}
-                      {%
-                        with
-                        label="Save and continue",
-                        name="save_and_continue",
-                        type="save"
-                      %}
-                        {% include "toolkit/button.html" %}
-                      {% endwith %}
+                      {{ govukButton({
+                        "text": "Save and continue",
+                        "name": "save_and_continue",
+                      }) }}
                     {% endblock %}
               </form>
               {%
--- a/app/templates/buyers/brief_overview.html
+++ b/app/templates/buyers/brief_overview.html
@@ -24,7 +24,10 @@
               with
               message = "Are you sure you want to delete these requirements?",
               type = "destructive",
-              action = '<button type="submit" class="button-destructive banner-action">Yes, delete</button>'|safe
+              action = govukButton({
+                "text": "Yes, delete",
+                "classes": "govuk-button--warning app-banner-action",
+              })
             %}
               {% include "toolkit/notification-banner.html" %}
             {% endwith %}
@@ -47,7 +50,10 @@
               heading = "Are you sure you want to withdraw these requirements?",
               banner_content = "<p>{}</p><ul><li>{}</li></ul>".format(withdraw_text, withdraw_list_items|join("</li><li>"))|safe,
               type = "destructive",
-              action = '<button type="submit" class="button-destructive banner-action">Withdraw requirements</button>'|safe
+              action = govukButton({
+                "text": "Withdraw requirements",
+                "classes": "govuk-button--warning app-banner-action",
+              })
             %}
               {% include "toolkit/notification-banner.html" %}
             {% endwith %}
--- a/app/assets/scss/application.scss
+++ b/app/assets/scss/application.scss
@@ -59,6 +59,9 @@ $govuk-compatibility-govukelements: true;
 @import "_secondary-action-link.scss";
 @import "_error-pages.scss";
 
+// Overrides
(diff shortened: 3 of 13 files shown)


risicle:
Ah, I've missed a couple...

risicle:
I can get buttons functioning correctly in-summary-table, but it looks like we're going to need a margin fix of some sort:

Before:
Spectacle X24158

after:
Spectacle L24158

app/templates/buyers/brief_overview.html
```diff
@@ -47,7 +47,7 @@
               heading = "Are you sure you want to withdraw these requirements?",
               banner_content = "<p>{}</p><ul><li>{}</li></ul>".format(withdraw_text, withdraw_list_items|join("</li><li>"))|safe,
               type = "destructive",
-              action = '<button type="submit" class="button-destructive banner-action">Withdraw requirements</button>'|safe
+              action = '<button class="govuk-button govuk-button--warning app-banner-action" data-module="govuk-button">Withdraw requirements</button>'|safe
```
aliuk2012: Can't we call the button component here instead of having a hardcoded button?

risicle: Yeah I guess - was just trying to make it as much a 1:1 translation as possible.

lfdebrux:
Nice work on getting the buttons working in the summary table.

Does the margin at the bottom have a big impact on usability? I'm wondering if we can leave it as is...

risicle:
Well, the only thing is it does add up when you have a large number of old briefs, meaning more scrolling. :man_shrugging:

risicle:
Cool - thanks to @aliuk2012 's advice I've got rid of the surplus margins.

lfdebrux:
> Cool - thanks to @aliuk2012 's advice I've got rid of the surplus margins.

Neat, what's the trick?

app/templates/buyers/award.html
```diff
@@ -44,14 +44,10 @@
...
-                        with
-                        label="Save and continue",
-                        name="save_and_continue",
-                        type="save"
-                      %}
-                        {% include "toolkit/button.html" %}
-                      {% endwith %}
+                      {{ govukButton({
+                        "text": "Save and continue",
+                        "name": "save_and_continue",
```
aliuk2012: why do we need the name? Does the backend need to know which button was clicked?

risicle:
> Neat, what's the trick?

`govuk-!-margin-0`

app/templates/buyers/award.html
risicle: Again, I'm just going for a 1:1 conversion. It's possible the analytics looks for it, it's possible the unit tests look for it, it's possible the FTs look for it...

risicle:
I also slightly wonder whether we should consider turning on double click protection for some of these buttons...

app/templates/buyers/award.html
aliuk2012: Could we remove and see if any breaks? Otherwise, it will always exist and we will always wonder why it's being used.

aliuk2012: Could we remove and see if any breaks? Otherwise, it will always exist and we will always wonder why it's being used.

risicle: Ok but we won't know for sure until it's got to the functional tests and even then I wouldn't be sure that e.g. our analytics code isn't trying to poke it... basically I'm afraid of scope creep here turning something that can be a straightforward task into something that requires much more thought and agreement - and we've got a long way to go on this task (I think this next round of apps are going to be significantly longer jobs than the previously converted frontends)

katstevens: I'd say leave it in - as Rob says it's the sort of thing that our event tracking analytics looks for.

aliuk2012: I'll not block this PR any further but I'm still a bit concerned that we have code which we are unsure of its purpose.

katstevens: Haha you wait until we get to the Supplier FE 😹

risicle: > a bit concerned that we have code which we are unsure of its purpose

"welcome to digital marketplace"

PR #237, opened 2019-11-07
DOS buyer review feature

katstevens:
https://trello.com/c/lvSNZWCR/64-3-create-single-tab-buyer-preview-page-on-briefs-fe

Allow buyers to see what their requirements will look like before they publish.

- I've called the feature 'review' instead of 'preview' because it's going to get very confusing otherwise
- Pulls in utils change to allow individual views to set `SAMEORIGIN` on the `X-Frame-Options` header
- Loads the `display_brief` content manifest in `app/main/__init__` to show the brief, as it's different from the `edit_brief` manifest, that the buyer uses to create the brief
- New views, templates and tests for the review page and the review 'source': 
  - both views should 400 if there are unanswered questions on the brief
  - all the `href` links in the source page should be disabled (they are set to `#` for now), with the exception of the GOV.UK logo, and the OGL license link.
  - Desktop and Mobile tabs seem to work!
  - some template elements will eventually be shared with the actual Buyer FE template that displays the published brief, but given most of the logic has been removed I'm 'duplicating' for now
- Added a comment on the `publish_brief` view because I spent at least 2 hours wondering if I needed to annotate my content sections with `questionAndAnswerSessionDetails` (answer: no)
- Content is all TBC, but can be changed in a separate PR. We need to decide whether to add `[DRAFT]` to the title etc to prevent confusion

goth-mobile-preview


tests/main/views/test_buyers.py
```diff
@@ -953,6 +953,298 @@ def test_404_if_question_does_not_exist(self):
...
+        assert len(document.xpath('//div[@class="govuk-tabs"]//a[contains(text(), "Desktop")]')) == 1
+
+        expected_src_link = "/buyers/frameworks/digital-outcomes-and-specialists-4/requirements/" \
+                            "digital-specialists/1234/review-source"
+        assert bool(document.xpath(
+            "//iframe[@src=$u][@title=$t][@class=$c]",
+            u=expected_src_link,
+            t="Preview of the page on desktop or tablet",
+            c="dm-desktop-iframe"
+        )) == 1
```
risicle: Wait wuh `bool(...) == 1`?

risicle:
Looks neat - I'll have to try it on monday.

tests/main/views/test_buyers.py
katstevens: It's true!

(Good spot - that should be a `len`)

app/templates/buyers/review_brief.html
```diff
@@ -0,0 +1,91 @@
+{% extends "_base_page.html" %}
+{% from "components/tabs/macro.njk" import govukTabs %}
```
aliuk2012: This might be better positioned in the `_base_page.html` so we know which components we are importing in one place and it might be used else where.

app/templates/buyers/review_brief.html
```diff
@@ -0,0 +1,91 @@
...
+      },
+      {
+          "link": url_for('external.list_opportunities', framework_family=brief.framework.family),
+          "label": "Supplier opportunities"
+      },
+      breadcrumbs
+    ]
+  %}
+    {% include "toolkit/breadcrumb.html" %}
+  {% endwith %}
```
aliuk2012: I know the breadcrumbs might not be in this application at the moment but maybe it might be good to get a start on it?

app/templates/buyers/review_brief.html
katstevens: I agree! Can you tell this branch was started before this repo had the new FE stuff pulled in? 😹

app/templates/buyers/review_brief.html
```diff
@@ -0,0 +1,91 @@
...
+
+{% block main_content %}
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <header class="page-heading-smaller">
+      <h1>Review your requirements</h1>
+    </header>
+  </div>
+</div>
```
aliuk2012: I think we could remove most of this and use only `h1` with govuk-frontend heading class to style the `h1`. As there is nothing else in the heading and it's short so we do not have to add all the extra markup for the grid.

app/templates/buyers/review_brief.html
```diff
@@ -0,0 +1,91 @@
...
+
+<div class="grid-row">
+  <div class="column-one-whole">
```
aliuk2012: Could we swap these out for govuk-frontend grids `govuk-grid-row` & `govuk-grid-column-full`?

Here is a really good reference to the classes we should be swapping to layout-and-grid-system-class-names

app/templates/buyers/review_brief_source.html
```diff
@@ -0,0 +1,170 @@
...
+</div>
+
+{% endblock %}
+{% block breadcrumb %}
+  {%
+    with
+    items = [
+      {
+          "link": "#",
+          "label": "Digital Marketplace"
```
aliuk2012: I was trying to understand why we were using so much of what is in the base page template. Is this because we wanted to remove URL's so users couldn't navigate away from the page in the iframe? My concern with this page at the moment is that we will need to remember about this other version of the base page when we come to replacing govuk_template with govuk-frontends template.

Have you thought about perhaps placing all the base page template stuff in another page under a `layout` folder with a name of say "no-navigation-links" and then just calling that page from here?

app/templates/buyers/review_brief_source.html
```diff
@@ -0,0 +1,170 @@
...
+            {{ (item_value | preserve_line_breaks) if item.type == 'textbox_large' else item_value }}
+        {% endwith %}
+      {% endcall %}
+    {% endcall %}
+{% endfor %}
+
+  </div>
+</div>
+<div class="grid-row">
+  <div class="column-one-whole">
```
aliuk2012: Could we use govuk-frontend classes?

app/templates/buyers/review_brief_source.html
```diff
@@ -0,0 +1,170 @@
...
+         "Completed applications",
+         "0 SME, 0 large",
+         id="completed-applications"
+       )
+    }}
+  </div>
+</div>
+
+<div class="grid-row">
+  <div class="column-one-whole">
```
aliuk2012: Could we use govuk-frontend classes?

app/templates/buyers/review_brief_source.html
```diff
@@ -0,0 +1,170 @@
...
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <header class="page-heading-smaller">
+      <p class="context">{{ brief.organisation }}</p>
+      <h1>{{ brief.title }}</h1>
+      <span data-lot="{{ brief.lotSlug }}"></span>
+    </header>
+  </div>
+</div>
```
aliuk2012: Could we use govuk-frontend classes?

katstevens: We can indeed - when this branch was started, they weren't available yet 😸

app/templates/buyers/review_brief.html
lfdebrux: I don't think we can use the breadcrumbs partially; the new style runs from the left margin of the page unless the base page layout is changed, and if you change the base page layout old style breadcrumbs get indented weirdly from where they should be.

app/templates/buyers/review_brief.html
katstevens: I can't get the spacing to work for this without also adding in the `_page_headings.scss` override? I think it might have to wait for now...

app/templates/buyers/review_brief.html
katstevens: Sure - the new classes seem to work ok

app/templates/buyers/review_brief_source.html
katstevens: I had an idea to share template components between the Buyer FE (where the template for the public page lives) and this page on the Briefs FE, and disable the links afterwards with Javascript (or something). Both apps still need FE work before this can feasibly be done. However this `review_brief_source.html` template is unlikely to be in this state for very long.

app/templates/buyers/review_brief_source.html
katstevens: Everything apart from the header spacing seems to work 👍

app/templates/buyers/review_brief.html
```diff
@@ -0,0 +1,90 @@
+{% extends "_base_page.html" %}
+
+{% block pageTitle %}{{ brief.title }} [DRAFT] - Digital Marketplace{% endblock %}
```
aliuk2012: For consistency 
```suggestion
{% block pageTitle %}
  {{ brief.title }} [DRAFT] - Digital Marketplace
{% endblock %}
```

app/templates/buyers/review_brief_source.html
```diff
@@ -0,0 +1,170 @@
+{% extends "_base_page.html" %}
+
+{% block page_title %}{{ brief.title }} [DRAFT] - Digital Marketplace{% endblock %}
```
aliuk2012:
```suggestion
{% block page_title %}
  {{ brief.title }} [DRAFT] - Digital Marketplace
{% endblock %}
```

risicle:
I realize that this is an MVP but d'y think we should put some big `DRAFT` notice on it somewhere in case someone accesses the view outside a frame and gets confused?

PR #238, opened 2019-11-08
Fix govuk-frontend assets path

lfdebrux:


katstevens:
We might need to updated the `$path` as well - just testing locally

katstevens:
Yep, we'll need to change

```
// Path to assets for use with file-url()
$path: "/buyers/static/images/";
```

lfdebrux:
Thanks Kat

PR #241, opened 2019-11-14
Replacing toolkit/page-headings with govuk frontend heading

aliuk2012:
This replaces toolkit/page-headings with govuk frontend page titles.

Generally, the headings were `h1` with a css class of `govuk-heading-xl`.
These were the mappings I used while replacing

Generally, use `govuk-heading-xl` class
If "smaller : true" when calling toolkit/page-heading, is `h1` with
`govuk-heading-l`.
If "Context" used when calling toolkit/page-heading, add span above the
`h1` and use govuk-caption-xl or govuk-caption-l depending on the heading
used.

ticket: https://trello.com/c/nxvnWqtc

### DIFF
--- a/app/templates/buyers/_base_edit_question_page.html
+++ b/app/templates/buyers/_base_edit_question_page.html
@@ -13,14 +13,9 @@
 
   <div class="grid-row">
     <div class="column-two-thirds">
-
-      {% with
-        heading = question.question,
-        smaller = true
-      %}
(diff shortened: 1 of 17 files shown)


lfdebrux:
Adding [WIP] until I get around to fixing the tests...

lfdebrux:
Tests should be fixed now

PR #249, opened 2019-11-29
Fixes malformed html link

aliuk2012:


lfdebrux:
While we're at it I think we can also remove the explanation list div and the large modifier from the paragraph

PR #250, opened 2019-12-04
Open DOS preview links in a new tab

katstevens:
https://trello.com/c/WSCRVNH8/149-allow-buyers-to-preview-external-links-on-their-brief

For the DOS preview feature, we want to allow buyers to check whether links in their brief content work OK. These links generally go to a 3rd party site (e.g. a public Google drive doc).

We already use a template filter `format_links` which picks up plain text URIs like `http://example.com` in user-generated content, and converts them to anchor links `<a href="http://example.com" class="break-link" rel="external">http://example.com</a>`. However in the DOS preview, these links would currently open within the iframe, and it's unlikely they would display correctly (e.g if the site being linked to has disabled the `X-Frame-Options` header - we can't control this). Therefore we want the links to open in a new tab.

In https://github.com/alphagov/digitalmarketplace-utils/pull/546 we modified `format_links` to accept a parameter `open_links_in_new_tab`. It seems to work! I've also tested that the published brief then opens the link in the current window, as per the current behaviour.

The [WCAG guidelines](https://www.w3.org/TR/WCAG20-TECHS/G200.html) say there should be a warning about any opening-in-new-tab behaviour. It doesn't feel suitable to put this next to the links in the preview pane itself (as suppliers won't see that message when it's published) so I've added a line of content above the panel.

### DIFF
--- a/app/templates/buyers/review_brief.html
+++ b/app/templates/buyers/review_brief.html
@@ -43,6 +43,7 @@
       {% endfor %}
     {% else %}
       <p class="govuk-body">This is how suppliers will see your requirements when they are published.</p>
+      <p class="govuk-body">Links will open in a new tab.</p>
       <p class="govuk-body">
         <a class="govuk-link" href="{{ url_for('.view_brief_overview', framework_slug=brief.frameworkSlug, lot_slug=brief.lotSlug, brief_id=brief.id) }}">Return to overview</a>
       </p>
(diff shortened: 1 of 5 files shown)


lfdebrux:
Good thinking about the content above the panel 👍

katstevens:
@lfdebrux Updated with the fixed version of utils.

PR #251, opened 2019-12-05
Add draft watermark to DOS Preview panels

katstevens:
https://trello.com/c/IwpYvT8z/148-add-draft-overlay-to-dos-preview-instead-of-draft

- Follows the GOV.UK Content Publisher pattern of using a watermark overlay on the draft preview.
- The ~CCS~ CSS is copied wholesale from the [GOV.UK static repo](https://github.com/alphagov/static/blob/6a0d8a619e56063c81c193458bff92618f163306/app/assets/stylesheets/helpers/_draft.scss). Most of the diff in this PR is whitespace changes, as I had to add a `draft-preview` wrapper div around the source page content. 
- I ran the Chrome dev tools accessibility audit and the text/background contrast passes 👍 
- I removed `[DRAFT]` from the brief title inside the panel, but kept it on the source page title in case anyone looks at the source page outside the iframe (unlikely?). 

Out of scope for this PR: navigation/buttons etc.

draft-requirements


risicle:
Oh yay we have the watermark

risicle:
> The CCS is copied wholesale

risicle:
> Links will open in a new tab.

Should this piece of content maybe say "Links _here_ will open in a new tab" or something to make it clear that it's referring to this page itself, not what will happen in the final brief?

katstevens:
That's a good question. As part of the navigation improvement ticket I'm going to add functional test support which should pick up any big howlers, but the long term plan was to have a shared toolkit template for both pages (though of course that relies on the Buyer FE and Briefs FE having the same version of the toolkit). This page hasn't changed much in a while, but if we do any significant improvements as part of the FE upgrade then we'll have to keep a close eye on it.

risicle:
Oh yes I remember you talking about this ... carry on.

katstevens:
Agree on the link content - I need to check the navigation content for the next ticket anyway, will make a note to review this as well.

app/assets/scss/_iframe_preview.scss
```diff
@@ -31,3 +31,10 @@ $app-mobile-border-bottom-width: 68px;
   height: 1000px;
   border: 0;
 }
+
+.draft-preview {
+  background-image: file-url("draft-watermark.png");
```
aliuk2012: This could have been a SVG or better yet it could have been text in a div and using CSS to rotate the text.  Both options could have scaled based on the size of the viewing point so it didn't pixelate.

katstevens: I copied it directly from the GOV.UK assets...

PR #252, opened 2019-12-09
Add call to action on DOS preview page

katstevens:
https://trello.com/c/HXr7GN2b/163-add-call-to-action-on-dos-preview-page

@NoraGDS and I agreed on some content changes for the top text, and wording for the action button at the bottom (that replaces the bottom 'Return to overview' link). 

I made the button part of a `GET` form, rather than a link that looks like a button (which we want to avoid). We follow this pattern already in the Briefs FE, when a buyer first starts a new brief.

Other stuff:
- Fixed the header styling to use the new GOVUK Frontend class (this page had fallen behind)
- Removed the confusing 'Apply for this opportunity' button inside the panel, so now there's only one call to action
- Removed footer links inside the panel. Ideally we'd remove the entire footer, but we'd have to pretty much clone the entire govuk base template to do that (suggestions on a non-annoying way to do this v welcome). 
- Rename stuff from 'review' to 'preview' as that's what everyone's been calling it. Recommend reviewing this commit separately!

What the button looks like:
confirm-button-footer


app/templates/buyers/preview_brief_source.html
```diff
@@ -54,11 +54,8 @@
 <div class="draft-preview">
   <div class="govuk-grid-row">
     <div class="column-two-thirds">
-      <header class="page-heading-smaller">
         <p class="context">{{ brief.organisation }}</p>
```
lfdebrux: To make this in line with GOV.UK Design System styles this should be a span

```suggestion
        <span class="govuk-caption-l">{{ brief.organisation }}</p>
```

PR #258, opened 2019-12-19
Bumps digitalmarketplace-govuk-frontend to 0.3.1

aliuk2012:
Ticket: https://trello.com/c/5aqSYf2P

### DIFF
--- a/config.py
+++ b/config.py
@@ -67,9 +67,13 @@ class Config(object):
     @staticmethod
     def init_app(app):
         repo_root = os.path.abspath(os.path.dirname(__file__))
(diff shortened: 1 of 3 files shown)


katstevens:
There's a failing test here because the 5xx error page is not rendering the expected error content, only the `<title>`. The response has

```
 <main id="content" role="main">
 </main>
```

which doesn't look like `/templates/toolkit/errors/500.html` to me...

lfdebrux:
We can't update the error templates until the breadcrumbs work has been completed. Going to close this PR for now.

PR #263, opened 2020-01-20
fixing breadcrumbs

agalamatis:
trello: https://trello.com/c/CO5wI047/179-3-update-breadcrumbs-to-use-govuk-frontend-component-in-briefs-frontend

### DIFF
--- a/app/templates/buyers/award.html
+++ b/app/templates/buyers/award.html
@@ -5,11 +5,34 @@
 {% endblock %}
 
 {% block breadcrumb %}
-  {%
-    with items = breadcrumbs
-  %}
-    {% include "toolkit/breadcrumb.html" %}
-  {% endwith %}
+  {{ govukBreadcrumbs({
+    "items": [
+        {
+          "href": "/",
+          "text": "Digital Marketplace"
+      },
+      {
+          "href": url_for("buyers.buyer_dashboard"),
+          "text": "Your account"
+      },
+      {
+          "href": url_for("buyers.buyer_dos_requirements"),
+          "text": "Your requirements"
+      },
+      {
+        "href": url_for(
+          ".view_brief_overview",
+          framework_slug=brief['frameworkSlug'],
+          lot_slug=brief['lotSlug'],
+          brief_id=brief['id']
+        ),
+        "text": brief['title']
+      },
+      {
+          "text": "Who won the contract?"
+      }
+    ]
+  }) }}
 {% endblock %}
 
 {% block mainContent %}
--- a/app/templates/buyers/award_details.html
+++ b/app/templates/buyers/award_details.html
@@ -7,11 +7,33 @@
 {% endblock %}
 
 {% block breadcrumb %}
-  {%
-    with items = breadcrumbs
-  %}
-    {% include "toolkit/breadcrumb.html" %}
-  {% endwith %}
+  {{ govukBreadcrumbs({
+    "items": [
+      {
+        "href": "/",
+        "text": "Digital Marketplace"
+      },
+      {
+        "href": url_for("buyers.buyer_dashboard"),
+        "text": "Your account"
+      },
+      {
+        "href": url_for("buyers.buyer_dos_requirements"),
+        "text": "Your requirements"
+      },
+      {
+        "href": url_for(
+          ".view_brief_overview",
+          framework_slug=brief['frameworkSlug'],
+          lot_slug=brief['lotSlug'],
+          brief_id=brief['id']),
+        "text": brief['title']
+      },
+      {
+        "text": "Tell us about your contract"
+      }
+    ]
+  }) }}
 {% endblock %}
 
 {% block mainContent %}
--- a/app/templates/buyers/award_or_cancel_brief.html
+++ b/app/templates/buyers/award_or_cancel_brief.html
@@ -7,9 +7,34 @@
 {% endblock %}
 
 {% block breadcrumb %}
-  {% with items = breadcrumbs %}
-    {% include "toolkit/breadcrumb.html" %}
-  {% endwith %}
+  {{ govukBreadcrumbs({
+    "items": [
+      {
+        "href": "/",
+        "text": "Digital Marketplace"
+      },
+      {
+        "href": url_for("buyers.buyer_dashboard"),
+        "text": "Your account"
+      },
+      {
+        "href": url_for("buyers.buyer_dos_requirements"),
+        "text": "Your requirements"
+      },
+      {
+        "href": url_for(
+          ".view_brief_overview",
+          framework_slug=brief['frameworkSlug'],
+          lot_slug=brief['lotSlug'],
+          brief_id=brief['id']
+        ),
+        "text": brief['title']
+      },
+      {
+          "text": "Have you awarded a contract?"
+      }
+    ]
+  }) }}
 {% endblock %}
 
 {% block mainContent %}
--- a/app/templates/buyers/brief_responses.html
+++ b/app/templates/buyers/brief_responses.html
@@ -4,22 +4,45 @@
   Responses to {{ brief.title or brief.lotName }} – Digital Marketplace
 {% endblock %}
 
+{% if response_counts['eligible'] > 0 %}
+  {% set heading = 'Shortlist suppliers' %}
+{% else %}
+  {% set heading = 'There were no applications' %}
+{% endif %}
+
 {% block breadcrumb %}
-  {%
-      with items = breadcrumbs
-  %}
-    {% include "toolkit/breadcrumb.html" %}
-  {% endwith %}
+  {{ govukBreadcrumbs({
+    "items": [
+        {
+          "href": "/",
+          "text": "Digital Marketplace"
+      },
+      {
+          "href": url_for("buyers.buyer_dashboard"),
+          "text": "Your account"
+      },
+      {
+          "href": url_for("buyers.buyer_dos_requirements"),
+          "text": "Your requirements"
+      },
+      {
+        "href": url_for(
+            ".view_brief_overview",
+            framework_slug=brief['frameworkSlug'],
+            lot_slug=brief['lotSlug'],
+            brief_id=brief['id']),
+        "text": brief['title']
+      },
+      {
+          "text": "Responses"
+      }
+    ]
+  }) }}
 {% endblock %}
 
 {% block mainContent %}
 <div class="grid-row">
     <div class="column-two-thirds">
-        {% if response_counts['eligible'] > 0 %}
-          {% set heading = 'Shortlist suppliers' %}
-        {% else %}
-          {% set heading = 'There were no applications' %}
-        {% endif %}
         <h1 class="govuk-heading-l">{{ heading }}</h1>
     </div>
 </div>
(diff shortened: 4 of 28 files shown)


app/templates/buyers/award.html
```diff
@@ -5,11 +5,34 @@
...
+        "href": url_for(
+          ".view_brief_overview",
+          framework_slug=brief['frameworkSlug'],
+          lot_slug=brief['lotSlug'],
+          brief_id=brief['id']
+        ),
+        "text": brief['title']
+      },
+      {
+          "text": "Who won the " + brief.title + " contract?"
```
katstevens: Let's keep it consistent within the breadcrumbs - either `brief.title` or `brief['title']` but not both.

app/templates/buyers/award_details.html
```diff
@@ -7,11 +7,33 @@
...
+      {
+        "href": url_for(
+          ".view_brief_overview",
+          framework_slug=brief['frameworkSlug'],
+          lot_slug=brief['lotSlug'],
+          brief_id=brief['id']),
+        "text": brief['title']
+      },
+      {
+        "text": "Tell us about your contract with " + pending_brief_response.supplierName
```
katstevens: Are we sure that `pending_brief_response.supplierName` will always be on the object? Worth having a default just in case (eg `the supplier`)?

app/templates/buyers/award_or_cancel_brief.html
```diff
@@ -7,9 +7,34 @@
...
+        "href": url_for(
+          ".view_brief_overview",
+          framework_slug=brief['frameworkSlug'],
+          lot_slug=brief['lotSlug'],
+          brief_id=brief['id']
+        ),
+        "text": brief['title']
+      },
+      {
+          "text": "Requirements already updated for " + brief.title
```
katstevens: This seems like an error message to me?

app/templates/buyers/award.html
```diff
@@ -5,11 +5,34 @@
...
 {% block breadcrumb %}
-  {%
-    with items = breadcrumbs
-  %}
-    {% include "toolkit/breadcrumb.html" %}
-  {% endwith %}
+  {{ govukBreadcrumbs({
+    "items": [
+        {
+          "href": "/",
```
katstevens: @lfdebrux Is there a reason we're not using `url_for('external.index')` in all of these? I can see `/` used on the other app templates as well...

lfdebrux: No reason that I'm aware of

app/templates/buyers/dashboard.html
```diff
@@ -6,11 +6,21 @@
...
+        "label": "Digital Marketplace"
+      },
+      {
+        "link": url_for("buyers.buyer_dashboard"),
+        "label": "Your account"
+      },
+      {
+        "text": "Your requirements"
+      }
+    ]
```
lfdebrux: Seems to be a problem with this one?

image

app/templates/buyers/award_details.html
agalamatis: if there is no brief_response the view throws a 404 and the DB the API endpoint hits has its supplier FK set as NOT NULL (https://github.com/alphagov/digitalmarketplace-api/blob/master/app/models/main.py#L1795), so the context variable shouldn't end up not having the supplier's name value.

app/templates/buyers/award_or_cancel_brief.html
agalamatis: Done.

app/templates/buyers/cancel_brief.html
```diff
@@ -7,11 +7,33 @@
...
+      {
+        "href": url_for(
+            ".view_brief_overview",
+            framework_slug=brief['frameworkSlug'],
+            lot_slug=brief['lotSlug'],
+            brief_id=brief['id']),
+        "text": brief['title']
+      },
+      {
+          "text": form.cancel_reason.label.text
```
lfdebrux: In the case when you're in the award flow, responding as to why you didn't award a brief, this is quite cluttered. It would be nice if the final breadcrumb read `Why didn't you award a contract?`, but that might be difficult to make happen given the way this view is constructed.

image

app/templates/buyers/award.html
agalamatis: @katstevens / @lfdebrux Should I change all of the slashes then? The slash is what I have seen used in almost all templates. Slashes would be faster and don't rely in the view's name being registered with the flask app.

app/templates/buyers/create_brief_question.html
```diff
@@ -11,20 +11,21 @@
...
+        "href": "/"
       },
       {
-        "link": url_for("dos.info_page_for_starting_a_brief", framework_slug=framework.slug, lot_slug=lot.slug),
-        "label": previous_title_text
+        "text": previous_title_text,
+        "href": url_for("dos.info_page_for_starting_a_brief", framework_slug=framework.slug, lot_slug=lot.slug)
+      },
+      {
+        "text": "Cancel brief question"
```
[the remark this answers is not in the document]
agalamatis: Yes, it was wrong, nice catch.

app/templates/buyers/award_or_cancel_brief.html
lfdebrux: This could be `Have you awarded a contract?`.

We don't need the requirement name in this final breadcrumb, as it is already in the previous breadcrumb.

image

app/templates/buyers/award.html
```diff
@@ -5,11 +5,34 @@
...
+        "href": url_for(
+          ".view_brief_overview",
+          framework_slug=brief['frameworkSlug'],
+          lot_slug=brief['lotSlug'],
+          brief_id=brief['id']
+        ),
+        "text": brief['title']
+      },
+      {
+          "text": "Who won the " + brief['title'] + " contract?"
```
lfdebrux: We can have this as just `Who won the contract?`

app/templates/buyers/award.html
agalamatis: Done.

app/templates/buyers/award_details.html
lfdebrux: We can have this as simply `Tell us about your contract`

app/templates/buyers/award_or_cancel_brief.html
agalamatis: Changed.

app/templates/buyers/brief_responses.html
```diff
@@ -4,22 +4,45 @@
...
+      {
+        "href": url_for(
+            ".view_brief_overview",
+            framework_slug=brief['frameworkSlug'],
+            lot_slug=brief['lotSlug'],
+            brief_id=brief['id']),
+        "text": brief['title']
+      },
+      {
+          "text": heading
```
lfdebrux: Can we change this to be `Responses`?

app/templates/buyers/dashboard.html
agalamatis: fixed with the incoming commit

app/templates/buyers/cancel_brief.html
agalamatis: Done too.

app/templates/buyers/award.html
agalamatis: Done too.

app/templates/buyers/index.html
```diff
@@ -5,16 +5,17 @@
...
-        "label": "Digital Marketplace"
-    }
+  {{ govukBreadcrumbs({
+    "items": [
+      {
+        "text": "Digital Marketplace",
+        "href": "/"
+      },
+      {
+        "text": current_user.name
```
lfdebrux: Could we have this as `Your account` to be consistent with breadcrumbs for later pages please?

app/templates/buyers/brief_responses.html
agalamatis: OK

app/templates/buyers/index.html
agalamatis: Changed.

app/templates/buyers/award_details.html
agalamatis: Changed.

app/templates/buyers/award.html
katstevens: Given that it's like this throughout the other apps, it might make sense to have a separate ticket to fix them all.

The `external.index` view name should already be available for all the FE flask apps to access.

agalamatis:
New commit with all the fixes (except the /), adjusted tests and squashed from master.

app/templates/buyers/create_brief_question.html
```diff
@@ -11,20 +11,21 @@
...
+        "href": "/"
       },
       {
-        "link": url_for("dos.info_page_for_starting_a_brief", framework_slug=framework.slug, lot_slug=lot.slug),
-        "label": previous_title_text
+        "text": previous_title_text,
+        "href": url_for("dos.info_page_for_starting_a_brief", framework_slug=framework.slug, lot_slug=lot.slug)
+      },
+      {
+        "text": question.question
```
lfdebrux: This seems wrong?

image

agalamatis: Changed to parent template's heading `question.question`.

app/templates/buyers/supplier_questions.html
```diff
@@ -5,11 +5,25 @@
...
+    "items": [
+      {
+        "href": "/",
+        "text": "Digital Marketplace"
+      },
+      {
+        "href": url_for("buyers.buyer_dashboard"),
+        "text": "Your account"
+      },
+      {
```
lfdebrux: I think this needs an additional breadcrumb to the requirement

image

app/templates/buyers/cancel_brief.html
lfdebrux: This doesn't work because the view and template are used for two different pages :(

https://github.com/alphagov/digitalmarketplace-briefs-frontend/blob/db93e3b565b035c224bbf118f52eeaa445d04713/app/main/views/outcome.py#L155-L240

It might be a pain to have different breadcrumbs to the page heading for this page...

image

agalamatis: @lfdebrux / @NoraGDS So use the heading? or the heading without the service name?

lfdebrux: If possible use the heading without the service name, otherwise just use the heading as before

app/templates/buyers/supplier_questions.html
agalamatis: @lfdebrux You mean to the following view? -> https://github.com/alphagov/digitalmarketplace-briefs-frontend/blob/master/app/main/views/buyers.py#L188

app/templates/buyers/cancel_brief.html
NoraGDS: In this case it would be: Why do you need to cancel?

app/templates/buyers/supplier_questions.html
lfdebrux: Yeah that's the one. I think other pages have it.

app/templates/buyers/cancel_brief.html
agalamatis: Thanks, coming in the next commit.

app/templates/buyers/supplier_questions.html
agalamatis: Done

PR #268, opened 2020-01-24
switch uses of grid & column classes to digitalmarketplace-govuk-frontend equivalents

risicle:
https://trello.com/c/ek1zy7Ce

Needed a couple of extended fixes to some table-based pages, helped by some advice from @lfdebrux . Other than that, I looked at all changed pages and they seemed to be visually identical.

### DIFF
--- a/app/templates/buyers/brief_overview.html
+++ b/app/templates/buyers/brief_overview.html
@@ -29,10 +29,10 @@
 {% endblock %}
 
 {% block mainContent %}
-  <div class="grid-row">
+  <div class="govuk-grid-row">
     {% block before_heading %}
       {% if delete_requested %}
-        <div class="column-one-whole">
+        <div class="govuk-grid-column-full">
           <form action="{{ url_for('buyers.delete_a_brief', framework_slug=framework.slug, lot_slug=brief.lotSlug, brief_id=brief.id) }}" method="POST">
             <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
             <input type="hidden" name="delete_confirmed" value="true" />
@@ -51,7 +51,7 @@
         </div>
       {% endif %}
       {% if withdraw_requested %}
-        <div class="column-one-whole">
+        <div class="govuk-grid-column-full">
           <form action="{{ url_for('buyers.withdraw_a_brief', framework_slug=framework.slug, lot_slug=brief.lotSlug, brief_id=brief.id) }}" method="POST">
             <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
             <input type="hidden" name="withdraw_confirmed" value="true" />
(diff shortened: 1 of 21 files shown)


risicle:
Oh forgot about the tests - will get back on this soon...

app/templates/buyers/brief_overview.html
```diff
@@ -15,10 +15,10 @@
 {% endblock %}
 
 {% block mainContent %}
-  <div class="grid-row">
+  <div class="govuk-grid-row">
     {% block before_heading %}
       {% if delete_requested %}
-        <div class="column-one-whole">
+        <div class="govuk-grid-column-one-whole">
```
lfdebrux: Looking at the [table of grid widths in govuk-frontend](https://github.com/alphagov/govuk-frontend/blob/v2.13.0/src/settings/_measurements.scss#L23-L30) I don't think `govuk-grid-column-one-whole` is included in govuk-frontend

lfdebrux:
We should now be able to remove some or all of the following Sass:

- [`_grid_layout.scss`](https://github.com/alphagov/digitalmarketplace-briefs-frontend/blob/1dfba29af508b41d1ad4fe02d2104b7b01441e6e/app/assets/scss/application.scss#L6)
- [`toolkit/_grids.scss`](https://github.com/alphagov/digitalmarketplace-briefs-frontend/blob/1dfba29af508b41d1ad4fe02d2104b7b01441e6e/app/assets/scss/application.scss#L27)
- [`overrides/_page-headings.scss`](https://github.com/alphagov/digitalmarketplace-briefs-frontend/blob/1dfba29af508b41d1ad4fe02d2104b7b01441e6e/app/assets/scss/application.scss#L65)

Have you tried this?

app/templates/buyers/brief_overview.html
risicle: Ah interesting so it's actually doing nothing, eh?

lfdebrux: Yeah I think so

lfdebrux: Should be replaced with `govuk-grid-column-full`

risicle: :+1: thanks

risicle:
> Have you tried this?

I have now. It appears to work ok, but I haven't gone through it as thoroughly as I did switching over the pages in the first place, because that takes a while...

lfdebrux:
> > Have you tried this?
> 
> I have now. It appears to work ok, but I haven't gone through it as thoroughly as I did switching over the pages in the first place, because that takes a while...

If you push up the change as a separate commit I'll check them over

risicle:
Done.

lfdebrux:
Hmm okay I think we should keep `overrides/_page-headings` in as it adds spacing between the page heading and the breadcrumbs that is currently missing, but will added once we replace govuk_template with govuk-frontend's template.

Otherwise these changes look good to me!

PR #280, opened 2020-02-19
Use pip-tools

lfdebrux:
Ticket: https://trello.com/c/iLQca47V/251-replace-custom-script-to-freeze-requirements-with-pip-tools

We want to remove our custom freeze-requirements script from dmutils; this PR replaces its use with pip-tools.


requirements.txt
```diff
@@ -1,59 +1,58 @@
...
-six==1.14.0
+pyyaml==5.3
+requests==2.22.0          # via mailchimp3, notifications-python-client
+s3transfer==0.3.3         # via boto3
+six==1.14.0               # via cryptography, python-dateutil
 unicodecsv==0.14.1
-urllib3==1.25.8
-Werkzeug==1.0.0
+urllib3==1.25.8           # via botocore, requests
+werkzeug==0.16.1          # via flask
```
katstevens: Back down to 0.16.1?

lfdebrux: Yeah, pip-compile is better about preventing/resolving conflicts than our old script. If we want to stay on Werkzeug 1.0.0 I need to push alphagov/digitalmarketplace-utils#556 to all the PRs

katstevens: Hmm OK. I don't think we're using any new 1.0.0 features yet.

lfdebrux: Not yet, but we might end up deleting some code because Werkzeug 1.x does URL slash merging!

PR #300, opened 2020-04-28
Update error messages for /buyer/create

domoscargin:
https://trello.com/c/Yy3pBfww/13-1-review-error-messages-for-create-a-buyer-account-journey-in-briefs-frontend

Update the form error messages on /buyer/create/<token> to follow [GOV.UK Design System guidance](https://design-system.service.gov.uk/components/error-message/#be-clear-and-concise).

### DIFF
--- a/app/create_buyer/forms/auth_forms.py
+++ b/app/create_buyer/forms/auth_forms.py
@@ -8,7 +8,7 @@ class EmailAddressForm(FlaskForm):
     email_address = DMStripWhitespaceStringField(
         "Your email address",
         validators=[
-            DataRequired(message="You must provide an email address"),
-            ValidEmailAddress(message="You must provide a valid email address"),
+            DataRequired(message="Enter an email address"),
+            ValidEmailAddress(message="Enter an email address in the correct format, like name@example.gov.uk"),
         ]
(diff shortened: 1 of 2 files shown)


app/create_buyer/forms/auth_forms.py
```diff
@@ -8,7 +8,7 @@ class EmailAddressForm(FlaskForm):
     email_address = DMStripWhitespaceStringField(
         "Your email address",
         validators=[
-            DataRequired(message="You must provide an email address"),
-            ValidEmailAddress(message="You must provide a valid email address"),
+            DataRequired(message="Enter an email address"),
+            ValidEmailAddress(message="Enter an email address in the correct format, like name@example.com"),
```
lfdebrux, line 12: I wonder if the example we provide should look like its from a public sector organisation, something like `name@gov.example` or `name@example.gov.uk` perhaps. Might be overthinking it though...

katstevens: I agree - buyers are unlikely to have `.com` emails

domoscargin: Done!

PR #307, opened 2020-05-07
Update tests to use DOS4

lfdebrux:
Requires alphagov/digitalmarketplace-test-utils#28

Apologies in advance to anyone who tries to review this.

### DIFF
--- a/tests/main/views/test_buyers.py
+++ b/tests/main/views/test_buyers.py
@@ -17,12 +17,12 @@ def find_briefs_mock():
     base_brief_values = {
         "createdAt": "2016-02-01T00:00:00.000000Z",
         "framework": {
-            "slug": "digital-outcomes-and-specialists-2",
-            "family": "digital-outcomes-and-specialists",
+            "slug": "digital-outcomes-and-specialists-4",
+            "family": "digital-outcomes-and-specialists-4",
             "status": "live",
-            "name": "Digital Outcomes and Specialists 2"
+            "name": "Digital Outcomes and Specialists 4"
         },
-        "frameworkSlug": "digital-outcomes-and-specialists-2",
+        "frameworkSlug": "digital-outcomes-and-specialists-4",
         "lot": "digital-specialists"
     }
 
@@ -260,7 +260,7 @@ class TestStartNewBrief(BaseApplicationTest):
         self.data_api_client = self.data_api_client_patch.start()
 
         self.data_api_client.get_framework.return_value = FrameworkStub(
-            slug='digital-outcomes-and-specialists',
+            slug='digital-outcomes-and-specialists-4',
             status='live',
             lots=[
                 LotStub(slug='digital-specialists', allows_brief=True).response()
@@ -279,7 +279,7 @@ class TestStartNewBrief(BaseApplicationTest):
 
     def test_404_if_lot_does_not_allow_brief(self):
         self.data_api_client.get_framework.return_value = FrameworkStub(
-            slug='digital-outcomes-and-specialists',
+            slug='digital-outcomes-and-specialists-4',
             status='live',
             lots=[
                 LotStub(slug='digital-specialists', allows_brief=False).response()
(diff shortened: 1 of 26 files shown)


tests/main/views/test_buyers.py
```diff
@@ -2345,7 +2425,7 @@ def test_404_if_brief_has_wrong_lot(self):
         assert res.status_code == 404
 
 
-class TestViewBriefResponsesPageForLegacyBrief(AbstractViewBriefResponsesPage):
+class TestViewBriefResponsesPageForLegacyBrief(BaseApplicationTest):
```
katstevens, line 2428: The `AbstractViewBriefResponsesPage ` base class contains shared tests that are run for this legacy class and the main one - while the users can no longer create legacy briefs, they can still view their old ones, so we need coverage for this.

tests/main/views/test_download_responses.py
```diff
@@ -413,7 +426,10 @@ class TestDownloadBriefResponsesCsv(BaseApplicationTest):
 
     def setup_method(self, method):
         super(TestDownloadBriefResponsesCsv, self).setup_method(method)
-        self.brief = BriefStub(status='closed').single_result_response()
+
+        # CSV download only works with DOS1
+        # TODO: Fix or remove this functionality?
```
katstevens, line 431: Buyers can still look at their old briefs. I'd keep these tests, and make it clear that we're using DOS1.

tests/main/views/test_outcome.py
```diff
@@ -343,8 +343,8 @@ def test_award_brief_details_post_raises_400_if_required_fields_not_filled(self)
         assert res.status_code == 400
         self._assert_masthead(document)
         error_spans = document.xpath('//span[@class="validation-message"]')
-        assert self._strip_whitespace(error_spans[0].text_content()) == "Youneedtoanswerthisquestion."
-        assert self._strip_whitespace(error_spans[1].text_content()) == "Youneedtoanswerthisquestion."
+        assert self._strip_whitespace(error_spans[0].text_content()) == "Enterthestartdate."
+        assert self._strip_whitespace(error_spans[1].text_content()) == "Enteracontractvalue."
```
katstevens, line 347: 🎉

tests/main/views/test_buyers.py
```diff
@@ -1886,15 +1886,15 @@ def test_show_draft_brief_summary_page(self, framework_status):
             'How to shortlist suppliers',
             'How to evaluate suppliers',
             'How to award a contract',
-            'Download the Digital Outcomes and Specialists contract',
+            'Download the Digital Outcomes and Specialists 4 contract',
```
katstevens, line 1889: These tests are failing because the `brief_overview.html` template uses `brief.frameworkName` instead of `brief.framework.name`.

lfdebrux: This test passes on my machine 🤔

katstevens: I'm getting the same errors as Travis when I run locally.

katstevens: (They pass when I change the template to use `framework.name`)

tests/main/views/test_download_responses.py
lfdebrux: 👍

tests/main/views/test_buyers.py
lfdebrux: The tests pass now that alphagov/digitalmarketplace-test-utils#28 has been merged. I can update the template anyway though.

katstevens: 🤔 Aha, running `make test` didn't update the test utils version for me locally, that's why.

tests/main/views/test_buyers.py
lfdebrux: Does this mean we need to run tests against both frameworks?

katstevens: No, only these ones in the legacy flow need DOS1 - the other journeys should be the same for all iterations.

lfdebrux: So for the legacy class we test against DOS1, and the main class we test against DOS4?

katstevens: That should work, yes - if we setup `self.framework_slug = ` for each class, then we can refer to that instead of hardcoding the framework on the shared tests and the class-specific tests.

lfdebrux: Done

PR #312, opened 2020-05-20
bump frameworks version

gidsg:
part of https://trello.com/c/6b806g3r/40-remove-map-pdf-from-dos-brief-where-supplier-will-work

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "dependencies": {
     "del": "^5.1.0",
-    "digitalmarketplace-frameworks": "https://github.com/alphagov/digitalmarketplace-frameworks.git#v17.9.0",
+    "digitalmarketplace-frameworks": "git+https://github.com/alphagov/digitalmarketplace-frameworks.git#v17.10.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
   },
   "dependencies": {
     "del": "^5.1.0",
-    "digitalmarketplace-frameworks": "https://github.com/alphagov/digitalmarketplace-frameworks.git#v17.9.0",
+    "digitalmarketplace-frameworks": "git+https://github.com/alphagov/digitalmarketplace-frameworks.git#v17.10.0",
```
lfdebrux, line 11: I don't think the `git+` is needed here?

```suggestion
    "digitalmarketplace-frameworks": "https://github.com/alphagov/digitalmarketplace-frameworks.git#v17.10.0",
```

gidsg: @lfdebrux think you're probably right but this is what got generated by running: 
```
npm install https://github.com/alphagov/digitalmarketplace-frameworks.git#v17.10.0 --save
``` 
so I assume this is OK, rather than doing the change by hand.

lfdebrux: 👍

PR #314, opened 2020-06-04
Use govukInput for buyer signup form

katstevens:
https://trello.com/c/Ht0k2BEJ/31-1-replace-text-boxes-with-govuk-frontend-text-input-component-in-create-a-buyer-account-journey

The Create Buyer Account form is relatively simple, consisting of just one text input. This PR directly replaces the WTForm-generated widget (which uses the old FE toolkit template) with the `govukInput` macro, following the steps in the User Frontend (see commit messages).

- Bumps govuk-frontend-jinja to fix the `indent.njk` errors
- Use the correct static path for this app
- Add in `govuk-label` override style, based on the User FE equivalent
- Replace the form input, using autocomplete settings in line with User FE and attributes from the existing WTForm

Tested the validation banners locally - seems to work ok!

Functional tests are OK up until the usual local Notify failure for this journey.

### DIFF
--- a/app/assets/scss/application.scss
+++ b/app/assets/scss/application.scss
@@ -53,6 +53,7 @@ $govuk-global-styles: true;
 
 // Overrides
 @import "overrides/_notifications_banner";
+@import "overrides/_govuk-label";
 
 // Misc styles
 // TODO: Move misc styling into their own partial files or the Digital Marketplace FE Toolkit
--- a/app/assets/scss/overrides/_govuk-label.scss
+++ b/app/assets/scss/overrides/_govuk-label.scss
@@ -1,0 +1,15 @@
+// We want to ensure that govuk-label components are consistent with existing
+// labels. To do this we change the default label class to be styled like a
+// medium heading. This style was copied from
+// https://github.com/alphagov/govuk-frontend/blob/v2.13.0/src/components/label/_label.scss#L27-L30
+//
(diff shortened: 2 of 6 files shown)


domoscargin:
Awesome!

I'm getting browser validation styles locally if I enter a malformed email address, instead of Error Messages.

Screenshot 2020-06-04 at 14 15 30

katstevens:
@domoscargin that's what we currently have on our User FE login form (where I copied this from... ) what's the recommended behaviour here? Let the browser do its thing or ensure our own error messages get shown?

domoscargin:
> @domoscargin that's what we currently have on our User FE login form (where I copied this from... ) what's the recommended behaviour here? Let the browser do its thing or ensure our own error messages get shown?

We should turn off HTML5 validation and display error messages: https://design-system.service.gov.uk/patterns/validation/.

That's what we're doing for this form on prod at the mo.

katstevens:
> > @domoscargin that's what we currently have on our User FE login form (where I copied this from... ) what's the recommended behaviour here? Let the browser do its thing or ensure our own error messages get shown?
> 
> We should turn off HTML5 validation and display error messages: https://design-system.service.gov.uk/patterns/validation/.

Interesting, so add `novalidate` to the form element?

We should do the same for the User FE form as well in that case.

domoscargin:
Yes, I think a `novalidate` will sort it - works for me locally for sure.

PR #315, opened 2020-06-05
[WIP] Move content from EmailForm class to template

phantas:
**Trello:** https://trello.com/c/N2q43lqd

**TODO:** error banner is defined on _base_page.html so this code doesn't work as is. An option would be to overwrite errors before including base_page.html

Screenshot 2020-06-05 at 18 41 49

Label works fine:

Screenshot 2020-06-05 at 18 43 32


phantas:
Not the direction we are going.

PR #324, opened 2020-06-18
Replace instruction list with Task List

domoscargin:
https://trello.com/c/SPGKu5rX/95-3-replace-instruction-list-with-govuk-design-system-task-list-pages-pattern-in-create-a-dos-opportunity-journey

Have implemented a task list that's as close to the [prototype kit](https://github.com/alphagov/govuk-prototype-kit/blob/master/docs/views/templates/task-list.html) as possible so that hopefully if/when the task list gets added to the Design System, it should be fairly easy to convert and remove the custom styling.

Pages seem to be working as expected and passing old tests.

### To do
There are several more states now, and this needs to be tested more thoroughly:

* Correct tags show - To do, in progress, Done, Cannot start yet and Optional - depending on state of work. I've started on testing these, but I'm sure there's an easier way to do it that I haven't figured out.
* Preview/Publish text behaviour (initially unclickable, clickable once all required questions are answered). This is related to the above - what's the easiest way of generating a brief that's ready to preview, without having to manually enter all the answers in the test?
* The new `section_status` object behaviour should be tested - maybe testing its effect on the page is enough?

### Issues
* The mix of task list and normal links might be a bit messy, after all. Once you're past the draft stage, I think there's only one further "Task" that we check the state of: "Let suppliers know the outcome". So that's currently set up as a task list item. Would setting ALL the links up as Task List items also be confusing since we're not checking the state of all of them? What tag would we give things that are just providing guidance?

Screenshot 2020-06-19 at 09 27 33

### DIFF
--- a/app/templates/buyers/brief_overview.html
+++ b/app/templates/buyers/brief_overview.html
@@ -157,15 +157,6 @@
                 'text': 'Awarded to ' + awarded_brief_response_supplier_name,
                 'allowed_statuses': ['awarded']
               },
-              'additional_instructions': [
-                {
-                  'href': url_for(".award_or_cancel_brief", framework_slug=brief.framework.slug, lot_slug=brief.lotSlug, brief_id=brief.id),
-                  'text': "Let suppliers know the outcome",
-                  'allowed_statuses': ['closed'],
-                  'startable': true,
-                  'active_tag': 'To do'
-                }
-              ],
               'links': [
                 {
                   'href': 'https://www.gov.uk/guidance/how-to-award-a-contract-when-you-buy-services',
@@ -176,6 +167,11 @@
                   'href': call_off_contract_url,
                   'text': "Download the " + brief.framework.name + " contract",
                   'allowed_statuses': ['draft', 'live', 'closed']
+                },
+                {
+                  'href': url_for(".award_or_cancel_brief", framework_slug=brief.framework.slug, lot_slug=brief.lotSlug, brief_id=brief.id),
+                  'text': "Let suppliers know the outcome",
+                  'allowed_statuses': ['closed']
                 }
               ]
             }
--- a/app/templates/buyers/brief_overview.html
+++ b/app/templates/buyers/brief_overview.html
@@ -37,8 +37,6 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
       {% block before_sections %}{% endblock %}
-      <ol class="instruction-list steps">
-
         {% with
           steps = [
             {
@@ -73,6 +71,7 @@
                 'cancelled': 'Done',
                 'unsuccessful': 'Done',
               },
+              'additional_instructions': publish_requirements_section_instructions,
               'links': publish_requirements_section_links
             },
             {
@@ -158,6 +157,15 @@
                 'text': 'Awarded to ' + awarded_brief_response_supplier_name,
                 'allowed_statuses': ['awarded']
               },
+              'additional_instructions': [
+                {
+                  'href': url_for(".award_or_cancel_brief", framework_slug=brief.framework.slug, lot_slug=brief.lotSlug, brief_id=brief.id),
+                  'text': "Let suppliers know the outcome",
+                  'allowed_statuses': ['closed'],
+                  'startable': true,
+                  'active_tag': 'To do'
+                }
+              ],
               'links': [
                 {
                   'href': 'https://www.gov.uk/guidance/how-to-award-a-contract-when-you-buy-services',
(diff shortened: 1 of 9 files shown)


lfdebrux:
It's definitely a messy page. What you've done is a step in the right direction though. Will discuss further offline.

lfdebrux:
A designer colleague has suggested we put the guidance links in the task list with a "Guidance" tag.

I think this is a great idea we should implement, going to push a commit doing so.

Screenshot showing a task list with one item, "How to shortlist suppliers", with a tag reading "Guidance"

app/templates/buyers/brief_overview.html
```diff
@@ -229,51 +228,82 @@ <h1 class="govuk-heading-xl">{{ brief.get('title', brief['lotName']) }}</h1>
...
+                      <a class="govuk-link" 
+                        href="{{ instruction.href }}"
+                        aria-describedby="step-{{ step_number }}-link-{{loop.index}}"
+                      >{{ instruction.text }}</a>
+                    </span>
+                    <strong class="govuk-tag govuk-tag--grey dm-task-list__tag" id="step-{{ step_number }}-instruction-{{ loop.index }}">{{ instruction.active_tag }}</strong>
+                  </li>
+                  {% else %}
+                  <li class="dm-task-list__item">
+                    <span>{{ instruction.text }}</span>
```
lfdebrux, line 295: Is there a reason why this `span` doesn't have the `dm-task-list__task-name` class?

domoscargin: There IS a reason, and the reason is I missed this...

tests/main/views/test_buyers.py
```diff
@@ -1922,21 +1922,41 @@ def test_show_draft_brief_summary_page(self, framework_status):
             )
         ]
 
-    @pytest.mark.parametrize('framework_status', ['live', 'expired'])
```
katstevens: We should keep coverage for both live and expired frameworks if we can - buyers do look at their old briefs, after all.

lfdebrux: This is about editing briefs though... can buyers edit their briefs from expired frameworks?

If a framework is expired, should we even show the tags? 🤔

katstevens: They can't edit them at all once they're published, regardless of framework status. But you can still evaluate a brief and update it with the outcome.

lfdebrux: Right. But I think this test is testing a brief that hasn't been published?

lfdebrux: So we need a new test?

lfdebrux: Yeah we need a new test

lfdebrux:
> A designer colleague has suggested we put the guidance links in the task list with a "Guidance" tag.
> 
> I think this is a great idea we should implement, going to push a commit doing so.
> 

This did not work in the end, because of the way guidance isn't the only thing that isn't a task in our task list.

See [my Trello comment on this for more](https://trello.com/c/SPGKu5rX/95-3-replace-instruction-list-with-govuk-design-system-task-list-pages-pattern-in-create-a-dos-opportunity-journey#comment-5f3aae6da852dc7962930caa).

lfdebrux:
I've finally finished with the tests for this, in the end I took the cheating way out and used snapshot testing.

domoscargin:
I can't approve my own PR, but this looks good to me as a stopgap before improving the flow of it entirely.

tests/main/views/test_requirement_task_list.py
```diff
@@ -8,323 +8,478 @@
 from ...helpers import BaseApplicationTest
 
 
-class TestBriefSummaryPage(BaseApplicationTest):
+class RequirementsTaskListPageTest:
```
katstevens, line 11: Tiny thing but can we call this class `TestRequirementsTaskListPage` instead please?

lfdebrux: We can't because then pytest would run it, and it's a base class, not a test suite on it's own. We could call it `BaseRequirementsTaskListPageTest` to make that clear?

katstevens: (Or if it's a shared test class, let's call it that.)

tests/main/views/test_requirement_task_list.py
```diff
@@ -8,323 +8,478 @@
...
+        """Test that each item has a tag with allowed text"""
+        tags = task_list_page.cssselect(self.task_list_tags_selector)
+
+        allowed_tags = ("Done", "To do", "In progress", "Optional", "Cannot start yet")
+
+        for tag in tags:
+            assert tag.text in allowed_tags
+
+
+class TestRequirementsTaskListPageDraftBrief(RequirementsTaskListPageTest, BaseApplicationTest):
```
katstevens, line 88: Grouping the tests by brief status is definitely a good idea 👍

tests/main/views/test_requirement_task_list.py
```diff
@@ -8,323 +8,478 @@
...
         self.data_api_client_patch.stop()
         super().teardown_method(method)
 
-    @staticmethod
-    def _get_links(document, xpath, text_only=None):
-        if text_only:
-            return [e.text_content() for e in document.xpath(xpath)]
-        return [
-            (e.text_content(), e.get('href')) for e in document.xpath(xpath)
+    def get_task_list_page(self, brief: dict) -> html.HtmlElement:
```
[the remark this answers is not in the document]
lfdebrux: I've added a message to the assert that might help a bit, here's an example of a failure:

```
tests/main/views/test_requirement_task_list.py E

============================================================== ERRORS ===============================================================
________ ERROR at setup of TestRequirementsTaskListPageLiveBrief.test_task_list_page_title_starts_with_brief_title[expired] _________

self = <tests.main.views.test_requirement_task_list.TestRequirementsTaskListPageLiveBrief object at 0x110c65208>
brief = {'applicationsClosedAt': '2016-04-07T00:00:00.000000Z', 'clarificationQuestions': [], 'clarificationQuestionsAreClosed': True, 'clarificationQuestionsClosedAt': '2016-04-02T00:00:00.000000Z', ...}

    @pytest.fixture
    def task_list_page(self, brief) -> html.HtmlElement:
>       return self.get_task_list_page(brief)

tests/main/views/test_requirement_task_list.py:44:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <tests.main.views.test_requirement_task_list.TestRequirementsTaskListPageLiveBrief object at 0x110c65208>
brief = {'applicationsClosedAt': '2016-04-07T00:00:00.000000Z', 'clarificationQuestions': [], 'clarificationQuestionsAreClosed': True, 'clarificationQuestionsClosedAt': '2016-04-02T00:00:00.000000Z', ...}

    def get_task_list_page(self, brief: dict) -> html.HtmlElement:
        url = f"/buyers/frameworks/{brief['framework']['slug']}/requirements/{brief['lotSlug']}/{brief['id']}"
        res = self.client.get(url)

>       assert res.status_code == 200, f"Cannot access requirements task list page for {brief['status']} brief at {url}"
E       AssertionError: Cannot access requirements task list page for live brief at /buyers/frameworks/digital-outcomes-and-specialists-4/requirements/digital-specialists/1234
E       assert 404 == 200
E        +  where 404 = <Response streamed [404 NOT FOUND]>.status_code

tests/main/views/test_requirement_task_list.py:38: AssertionError
```

I understand what you're saying about wanting to understand a test in isolation, but if we take that principle to its logical end we end up setting up the framework and brief for each test, and I feel most of the actually useful information about what is supposed to be tested gets lost in setup. That's just like, my personal opinion though.

I think `pytest` is good at giving you the pertinent information though. There wouldn't be much difference in the error message between having the assert in the test fixture or the test itself.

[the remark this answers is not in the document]
lfdebrux: What about adding this test to the shared test class?

```
def test_can_view_task_list_page(self, brief):
        assert self.get_task_list_page(brief)
```

tests/main/views/test_requirement_task_list.py
```diff
@@ -8,323 +8,478 @@
...
-            'How to answer supplier questions',
-            'How to shortlist suppliers',
-            'How to evaluate suppliers',
-            'How to award a contract',
-            'Download the Digital Outcomes and Specialists 4 contract',
-        ]
+    def test_draft_brief_requirements_task_list(self, task_list_html, snapshot):
+        assert task_list_html == snapshot
+
+    def test_draft_brief_can_be_deleted(self, task_list_page):
```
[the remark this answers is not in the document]
lfdebrux: Done

tests/main/views/test_requirement_task_list.py
lfdebrux: Done

tests/main/views/test_requirement_task_list.py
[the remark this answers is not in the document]
katstevens: OK, that sounds reasonable 👍 

> What about adding this test to the shared test class?
Yep!

(Did you push recent changes by the way? I can't see them...)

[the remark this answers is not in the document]
lfdebrux: D'oh 🤦 Thanks for the heads up. Pushed now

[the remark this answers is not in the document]
katstevens: Weird - I still can't see them (on github or local branch update)?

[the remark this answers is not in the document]
lfdebrux: Ah, I forgot to commit the changes... can you see them now?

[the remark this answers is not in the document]
katstevens: There they are! 😸 LGTM

PR #329, opened 2020-07-17
Reorganise the view functions in app.main.views.buyers

lfdebrux:
I was having a terrible time trying to write tests for #324 because the file `tests/main/views/test_buyers.py` is so damn big, and talking with Kat I realised that we should just split the file up; so this PR does that :)

To make sure that the test files still align with the view files, I've also split up the file `app/main/views/test_buyers.py`, so now:

- `app.main.views.buyers` should contain views for the buyer dashboards
- `app.main.views.create_a_requirement.*` contains the views for the various stages of creating a requirement, up to publishing (and possibly withdrawing)
- `app.main.views.requirement_task_list` contains just the view for the task list for the requirement lifecycle

So this PR increases the number of files and the number of lines in the repo, but the files are on average shorter, and `test_buyers.py` is now less of a hellhole! :tada:

---

Thoughts welcome


lfdebrux:
2,551 lines removed from tests/main/views/test_buyers.py

😂

katstevens:
Epic! I haven't gone through in detail yet, but a few initial thoughts:

- I don't think the `withdraw` action should be in the `create_a_requirement` subfolder. It's quite different from deleting a draft brief - the brief must be live, and it doesn't disappear from public view (it just has the blue banner). I'd probably either put it in a separate file at `views.withdraw`, or keep in `buyers.py` for now.
- I'm not sure about using the term `requirement` in our file/folder names. I know that's what buyers see, but we use 'brief' pretty much everywhere in the code. It's also shorter 😹 
- Great catch on the `/copy` view. I didn't realise the edit view could potentially clash with it. And a nice explanatory commit message too 🏆!

lfdebrux:
> * I don't think the `withdraw` action should be in the `create_a_requirement` subfolder. It's quite different from deleting a draft brief - the brief must be live, and it doesn't disappear from public view (it just has the blue banner). I'd probably either put it in a separate file at `views.withdraw`, or keep in `buyers.py` for now.

:+1:

> * I'm not sure about using the term `requirement` in our file/folder names. I know that's what buyers see, but we use 'brief' pretty much everywhere in the code. It's also shorter 😹

I realise that we use brief everywhere, but I personally think that we should be trying to make our code/model consistent with the language we use with buyers; in Domain Driven Design there is a concept of having a [ubiquitous language](https://martinfowler.com/bliki/UbiquitousLanguage.html) which is a concept I think we should be looking at. Also I was thinking it would be cool if the files in our apps were structured in a way that reflects the journeys users follow, to make it easier for people to guess where to find things!

katstevens:
> we should be trying to make our code/model consistent with the language we use with buyers

I broadly agree. However it's probably not for this PR? I think it might cause more confusion than it solves, especially as we're unlikely to change all our model/view/template names in the immediate future.

lfdebrux:
> > we should be trying to make our code/model consistent with the language we use with buyers
> 
> I broadly agree. However it's probably not for this PR? I think it might cause more confusion than it solves, especially as we're unlikely to change all our model/view/template names in the immediate future.

Done

app/main/views/withdraw_brief.py
```diff
@@ -0,0 +1,32 @@
...
+
+from app import data_api_client
+from .. import main
+from ..helpers.buyers_helpers import (
+    get_framework_and_lot,
+    is_brief_correct,
+)
+
+
+BRIEF_WITHDRAWN_MESSAGE = "You’ve withdrawn your requirements for ‘{brief[title]}’"
```
katstevens, line 12: This line is still in `views.buyers`, I think

lfdebrux: Good spot!

PR #337, opened 2020-08-11
Use govukInput for text questions

lfdebrux:
https://trello.com/c/YXsBOrDK/110-3-replace-text-boxes-with-govuk-frontend-text-input-component-in-create-a-dos-opportunity-journey

Uses alphagov/digitalmarketplace-content-loader#90 to render questions with `type: text` with GOV.UK Design System text input components.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -3,6 +3,7 @@ from flask_login import LoginManager
 from flask_wtf.csrf import CSRFProtect
 
 import dmapiclient
+import dmcontent.govuk_frontend
 from dmutils import init_app
 from dmutils.user import User
 
@@ -52,6 +53,11 @@ def create_app(config_name):
     login_manager.login_message = None  # don't flash message to user
     gds_metrics.init_app(application)
     csrf.init_app(application)
+
+    # We want to be able to access this function from within all templates
+    application.jinja_env.globals["govuk_frontend_from_question"] = (
+        dmcontent.govuk_frontend.from_question
+    )
 
     @application.before_request
     def remove_trailing_slash():
--- a/app/assets/scss/overrides/_govuk-label.scss
+++ b/app/assets/scss/overrides/_govuk-label.scss
@@ -13,3 +13,8 @@
 
   margin-bottom: govuk-spacing(2);
 }
+
+.govuk-label--l {
+  @include govuk-font($size: 36, $weight: bold);
+  margin-bottom: govuk-spacing(3);
+}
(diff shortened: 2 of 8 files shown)


app/assets/scss/overrides/_govuk-label.scss
```diff
@@ -13,3 +13,8 @@
 
   margin-bottom: govuk-spacing(2);
 }
+
+.govuk-label--l {
```
domoscargin, line 17: Yup, the `.govuk-label` override takes precedence over previous styling, so is ignoring the `.govuk-label--l` styles before it.

app/__init__.py
```diff
@@ -53,6 +54,11 @@ def create_app(config_name):
     gds_metrics.init_app(application)
     csrf.init_app(application)
 
+    # We want to be able to access this function from within all templates
+    application.jinja_env.globals["govuk_frontend_from_question"] = (
+        dmcontent.govuk_frontend.from_question
+    )
```
katstevens, line 60: 🧙 Wizardry!

app/templates/buyers/_base_edit_question_page.html
```diff
@@ -7,24 +7,32 @@
 
 {% block mainContent %}
 
+  {% set is_govuk_frontend = govuk_frontend_from_question(question) %}
```
katstevens, line 10: So at the moment, this will only be truthy if `question` is type `text`, correct?

app/templates/buyers/_base_edit_question_page.html
```diff
@@ -7,24 +7,32 @@
...
   <form method="post" enctype="multipart/form-data" action="{{ request.path }}">
 
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-two-thirds">
           {% if question.type != 'multiquestion' %}
-            {% if errors and errors[question.id] %}
+            {% if is_govuk_frontend %}
+              {% set macro_name, params = govuk_frontend_from_question(question, brief, errors) %}
+              {% set govuk_forms = {"govukInput": govukInput} %}
+              {{ govuk_forms[macro_name](params) }}
```
katstevens, line 34: This is so lovely and neat 👍

app/templates/buyers/_base_edit_question_page.html
lfdebrux: Correct

app/templates/buyers/_base_edit_question_page.html
lfdebrux: It's going to get more complicated 😅

PR #339, opened 2020-08-14
Replace list form with List Input in Create a DOS Opportunity journey

domoscargin:
https://trello.com/c/VG4caYdJ/136-3-replace-list-form-with-list-input-component-in-create-a-dos-opportunity-journey

This PR uses the new [content-loader govuk_frontend methods](https://github.com/alphagov/digitalmarketplace-content-loader/pull/90) to use the [List Input component](https://github.com/alphagov/digitalmarketplace-govuk-frontend/tree/master/src/digitalmarketplace/components/list-input).

### DIFF
--- a/app/templates/buyers/_base_edit_question_page.html
+++ b/app/templates/buyers/_base_edit_question_page.html
@@ -4,19 +4,36 @@
 {% from "govuk/components/label/macro.njk" import govukLabel %}
 {% from "digitalmarketplace/components/list-input/macro.njk" import dmListInput %}
 
+{%- macro show_question(is_govuk_frontend, question, brief, errors, is_only_question=True) %}
+  {% if is_govuk_frontend %}
+    {% set form = govuk_frontend_from_question(question, brief, errors, is_page_heading=is_only_question) %}
+    {% set govuk_forms = {"govukInput": govukInput, "dmListInput": dmListInput} %}
+    {%- if form.label %}
+      {{ govukLabel(form.label) }}
+    {% endif -%}
+    {% if not form.params.question_advice %}
+      {{ question.question_advice }}
+    {% endif %}
+    {{ govuk_forms[form.macro_name](form.params) }}
+  {% elif errors and errors[question.id] %}
+    {{ forms[question.type](question, brief, errors) }}
+  {% else %}
+    {{ forms[question.type](question, brief, {}) }}
+  {% endif %}
+{% endmacro -%}
+
 {% block pageTitle %}
   {{ question.question }} – Digital Marketplace
 {% endblock %}
 
 {% block mainContent %}
-
   {% set is_govuk_frontend = govuk_frontend_from_question(question) %}
 
   {% if question.type != 'multiquestion' %}
     <div class="single-question-page">
   {% endif %}
 
-  {% if not is_govuk_frontend %}
+  {% if not is_govuk_frontend or question.type == 'multiquestion' %}
   <div class="govuk-grid-row">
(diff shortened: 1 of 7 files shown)


app/templates/buyers/_base_edit_question_page.html
```diff
@@ -31,10 +32,14 @@ <h1 class="govuk-heading-l">
...
           {% if question.type != 'multiquestion' %}
             {% if is_govuk_frontend %}
-              {% set govuk_forms = {"govukInput": govukInput} %}
+              {% set govuk_forms = {"govukInput": govukInput, "dmListInput": dmListInput} %}
               {% set form = govuk_frontend_from_question(question, brief, errors) %}
-              {{ govukLabel(form.label) }}
-              {{ question.question_advice }}
+              {%- if form.label %}
+                {{ govukLabel(form.label) }}
+              {% endif -%}
```
lfdebrux, line 39: 👍

lfdebrux:
Testing this locally, I've noticed that the new version is being used for the cultural fit criteria question but not the technical competence criteria question. I'm guessing this is because the latter is a multiquestion; shall we try and crack that problem in this PR? I'm happy to contribute a commit or pair on fixing that.

domoscargin:
> Testing this locally, I've noticed that the new version is being used for the cultural fit criteria question but not the technical competence criteria question. I'm guessing this is because the latter is a multiquestion; shall we try and crack that problem in this PR? I'm happy to contribute a commit or pair on fixing that.

That's a good point. I'd assumed we were handling multiquestions in content loader, but the logic does all appear to be contained here. I've updated the template page. The only issue is headings - we need Content Loader to not set individual question labels/legends as the page heading.

app/templates/buyers/_base_edit_question_page.html
```diff
@@ -53,7 +53,18 @@ <h1 class="govuk-heading-l">
...
+                {% set govuk_forms = {"govukInput": govukInput, "dmListInput": dmListInput} %}
+                {% set form = govuk_frontend_from_question(question, brief, errors) %}
+                {%- if form.label %}
+                  {{ govukLabel(form.label) }}
+                {% endif -%}
+                {% if not form.params.question_advice %}
+                  {{ question.question_advice }}
+                {% endif %}
+                {{ govuk_forms[form.macro_name](form.params) }}
+              {% elif errors and errors[question.id] %}
```
lfdebrux, line 67: Let's encapsulate this logic common to multiquestions and questions in a macro, we can define one at the top of this file https://jinja.palletsprojects.com/en/2.11.x/templates/#macros

lfdebrux:
> Guess we could either pass another arg to `govuk_frontend_from_question` indicating that it's a child, or we could handle multiquestions within Content Loader.

Let's add an optional parameter to `govuk_label` and `govuk_fieldset` `is_page_heading=True`, and also have `from_question` pass on kwargs. We might want to get more sophisticated later but this seems like a good way to start to me.

lfdebrux:
> > Guess we could either pass another arg to `govuk_frontend_from_question` indicating that it's a child, or we could handle multiquestions within Content Loader.
> 
> Let's add an optional parameter to `govuk_label` and `govuk_fieldset` `is_page_heading=True`, and also have `from_question` pass on kwargs. We might want to get more sophisticated later but this seems like a good way to start to me.

Done in [digitalmarketplace-content-loader#7.14.0](https://github.com/alphagov/digitalmarketplace-content-loader/pull/94)

lfdebrux (approved):
Looks great to me! Well done for getting this working 👍

PR #341, opened 2020-08-19
Fix focus style for links in task list

lfdebrux:
Previously the focus style extended leftward for a bit.

### Before

focus style extending leftwards

### After

focus style limited to link text


domoscargin (approved):
Much nicer!

PR #348, opened 2020-08-26
Replace WTForms toolkit elements with govuk-frontend components on DOS Opportunity journey

domoscargin:
https://trello.com/c/yHSxrqS9/173-3-replace-wtforms-toolkit-elements-with-govuk-frontend-components-in-award-a-dos-opportunity-journey

I've left the WTForms classes as they are and mostly focused on the view/template here.

I've had to hard-code into the template rather than use components for a couple of the pages because we have paragraphs between the heading and the radios that can't be handled by the GOVUK Frontend components, but we also want everything to be linked as a fieldset.

Currently running functional tests locally.

### DIFF
--- a/app/main/views/outcome.py
+++ b/app/main/views/outcome.py
@@ -16,7 +16,7 @@ from ..forms.award_or_cancel import AwardOrCancelBriefForm
 from dmapiclient import HTTPError
 from dmutils.flask import timed_render_template as render_template
 from dmutils.forms.errors import govuk_errors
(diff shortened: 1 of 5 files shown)


katstevens (approved):
All looks good to me 🏆 

I've only now realised that the `award_brief_details` view in this journey is using a Content Loader form! According to the git blame it was me who did this, 3 years ago 😹  Are we going to tackle that one separately?

PR #349, opened 2020-08-27
Use govuk frontend components on Award Details page

domoscargin:
@katstevens pointed out this page was using content-loader but hadn't yet been switched over to govuk frontend components, so this PR does that.

I've pulled out the macro which was on `base_edit_question_page` so we can use it in multiple spots.

Related PR:
https://github.com/alphagov/digitalmarketplace-content-loader/pull/100


katstevens (approved):
WFM!

PR #353, opened 2020-09-07
Migrate Briefs to GOVUK Frontend V3

domoscargin:
https://trello.com/c/VhmOUlKs/211-3-migrate-briefs-frontend-to-govuk-frontend-v3

This PR updates GOVUK Frontend to version 3 and removes legacy dependencies and assets.

PR #354, opened 2020-09-07
Replace alphagov/govuk-frontend-jinja with LandRegistry/govuk-frontend-jinja

lfdebrux:
We want to use the Jinja govuk-frontend templates built by the HMLR team at https://github.com/LandRegistry/govuk-frontend-jinja.

There are a couple of workarounds/tweaks needed to get it all working:

- We need to upgrade Jinja2 to 2.11 so we can use ChainableUndefined, otherwise attribute accesses such as `errors.email_address.errorMessage` will raise an exception
- We need to do add LandRegistry/govuk_frontend_jinja to the loader twice, and workaround the fact that its filenames end with `.html` instead of `.njk`

Also, we shouldn't deploy this to production, because LandRegistry/govuk_frontend_jinja assumes govuk-frontend v3 assets, so JavaScript features such as error summary links won't work (as the `data-` attribute values change between v2 and v3).

This PR should help keep the complexity of #353 down though.

### DIFF
--- a/config.py
+++ b/config.py
@@ -67,9 +67,13 @@ class Config(object):
 
     @staticmethod
     def init_app(app):
+        class GOVUKFrontendJinjaLoader(jinja2.PackageLoader):
+            # work around the fact that the filenames are different in govuk_frontend_jinja
+            def get_source(self, environment, template):
+                return super().get_source(environment, template.replace(".njk", ".html"))
+
         repo_root = os.path.abspath(os.path.dirname(__file__))
         digitalmarketplace_govuk_frontend = os.path.join(repo_root, "node_modules", "digitalmarketplace-govuk-frontend")
-        govuk_frontend = os.path.join(repo_root, "node_modules", "govuk-frontend")
         template_folders = [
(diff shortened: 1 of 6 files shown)


domoscargin (approved):
🥇 

LGTM

config.py
```diff
@@ -67,17 +67,24 @@ class Config(object):
 
     @staticmethod
     def init_app(app):
```
katstevens, line 69: I can't remember why we have this static method in `config.py` rather than say, `app/__init__.py`, or how it gets called? It'd be nice to keep `config.py` for well, config (not code). I don't know how difficult it is to move the stuff below somewhere else? For now, could we add a comment?

lfdebrux: Let me try moving it!

lfdebrux: Seems to work either way. Looking at the git log it isn't clear whether it was important for it to be in config.py or not.

katstevens: Yeah looking at the `git blame`, the method seems to have been there since the dawn of time 😸 I think it's just more noticeable the more complex it gets. Maybe leave for tech debt?

PR #356, opened 2020-09-08
Revert "Replace alphagov/govuk-frontend-jinja with LandRegistry/govuk-frontend-jinja"

lfdebrux:
Reverts alphagov/digitalmarketplace-briefs-frontend#354

I think this is causing 502s on preview, having issues locally.


agalamatis (approved):
Let's give it a try.

PR #358, opened 2020-09-09
Add test for function of dmAlert for flash messages

lfdebrux:
While trying to migrate this app to LandRegistry/govuk-frontend-jinja I got caught out by an issue with the `dmAlert` component that wasn't tested by our unit tests but was tested by our functional tests. We should test that the template works in our unit tests.


domoscargin (approved):
Looks good and passing locally for me.

lfdebrux:
Based on a suggestion by @katstevens I've changed this so that the test of the display of the flash message is a separate method to just testing if the flash message is present in the sesssion.

PR #360, opened 2020-09-09
Fix link without class

domoscargin:
In v3, this unclassed link uses browser default styles.

So we just add the `govuk-link` class.


lfdebrux (approved):
Good spot!

PR #383, opened 2021-01-29
Use render_question() to get HTML for multiquestion forms

lfdebrux:
Depends on alphagov/digitalmarketplace-content-loader#130.

digitalmarketplace-content-loader>=7.29.0 supports multiquestions, we can get rid of the special logic in `_base_edit_question_page.html`.

### DIFF
--- a/app/templates/buyers/_base_edit_question_page.html
+++ b/app/templates/buyers/_base_edit_question_page.html
@@ -20,18 +20,7 @@
 
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-two-thirds">
-          {% if question.type != 'multiquestion' %}
-            {{ render_question(question, brief, errors) }}
-          {% else %}
-            {% if question.question_advice %}
-              <span class="question-advice">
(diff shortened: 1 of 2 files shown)


app/templates/buyers/_base_edit_question_page.html
```diff
@@ -20,18 +20,7 @@ <h1 class="govuk-heading-l">
 
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-two-thirds">
-          {% if question.type != 'multiquestion' %}
-            {{ render_question(question, brief, errors) }}
-          {% else %}
-            {% if question.question_advice %}
-              <span class="question-advice">
```
alex9smith, line 27: Do we need any CSS updates for the new `dm-question-advice` class?

alex9smith: Good, I couldn't see it anywhere either but wanted to check

alex9smith (approved):
👍

PR #419, opened 2021-04-07
prevent 500 when viewing timeline after publishing brief

gidsg:
https://trello.com/c/jOWgypcV/2197-500-error-on-https-wwwdigitalmarketplaceservicegovuk-buyers-frameworks-digital-outcomes-and-specialists-5-requirements-digital-o
Previously we didn't handle the case where the buyer viewed the timeline after publishing an opportunity with a question and answer session. This was previously causing a 500 as the behaviour for the condition was not specified. 

The intended behaviour in this case; as confirmed by @NoraGDS; is that they can still see the question and answer session on the timeline after publishing (but without the Edit link). This is what is implemented by this PR.

Screenshot 2021-04-07 at 14 24 49


bjgill (commented):
Looks good. Is there a test for this to prevent future regression?

gidsg:
> Is there a test for this to prevent future regression?

@bjgill good point. I added a test now in 7c09283.

bjgill (approved):
:shipit:

PR #427, opened 2021-04-19
Give me access to download a brief responses file

bjgill:
https://govuk.zendesk.com/agent/tickets/4533642

I'm investigating a report of the downloaded file having been mangled - its contents don't match what's in the database. I want to narrow down the possibilities for how it happened by downloading the file myself to see if I also see the same mangling. We don't support adding another user to a brief, so I'm using a rather ugly hack to give myself this temporary access.

This change must be reverted once no longer needed, and certainly no later than 23 April 2021.

### DIFF
--- a/app/main/helpers/buyers_helpers.py
+++ b/app/main/helpers/buyers_helpers.py
@@ -26,6 +26,25 @@ def is_brief_correct(brief, framework_slug, lot_slug, current_user_id, allow_wit
     )
 
 
+def is_brief_correct_debug(
+        brief, framework_slug, lot_slug, current_user_id, allow_withdrawn=False, allowed_statuses=None
+):
+    """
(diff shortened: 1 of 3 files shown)


alex9smith (approved):
I don't like this very much, but I can't think of a better way to debug this issue given we don't have the brief responses in our other environments.

alex9smith:
Oh, just saw the test failures. @bjgill That might not be simple to resolve. Would it be possible to create a script that copies the brief responses from prod onto the brief on staging instead?

bjgill:
I agree - it's an ugly hack. Just force-pushed a test fix. I'll merge once the tests are clean and try to get it reverted as soon as I can.

I'll create a story to think about how we can do this sort of debugging with less hacking in future.

PR #428, opened 2021-04-20
Revert "Give me access to download a brief responses file"

bjgill:
https://govuk.zendesk.com/agent/tickets/4533642

I've downloaded the response file in question and got what I needed for my investigation. We should now revert this to remove my hacky access.

Reverts alphagov/digitalmarketplace-briefs-frontend#427


alex9smith (approved):
🔥

PR #446, opened 2021-06-08
Rename labs to studios

gidsg:
https://trello.com/c/ohDPcudm/606-1-change-dos-page-title-user-research-studios
We want to rename user research labs to studios to reflect the current name of the lot.

### DIFF
--- a/app/templates/buyers/studios_start_page.html
+++ b/app/templates/buyers/studios_start_page.html
@@ -65,7 +65,7 @@
       <li>Award a contract to the supplier that best meets your needs.</li>
     </ol>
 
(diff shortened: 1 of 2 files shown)


app/templates/buyers/studios_start_page.html
```diff
@@ -67,11 +67,11 @@ <h2 class="govuk-heading-m">Evaluate</h2>
 
     <p class="govuk-body">Read more about <a class="govuk-link" href="https://www.gov.uk/guidance/how-to-hire-user-research-labs-on-the-digital-marketplace">how to hire user research labs</a>.</p>
```
bjgill, line 68: Is this `labs` being left as-is deliberately?

gidsg: Yes @bjgill that URL is on GOV.UK so not something we control directly (especially having no content designer currently).

gidsg: fixed link text in 7a96ca9, we cannot control the link URL as that is on GOV.UK.

PR #452, opened 2021-06-15
Get sourcemaps from gulp

bjgill:
https://trello.com/c/zjkM8Utr/8-3-upgrade-off-gulp-sourcemaps-v3

As of Gulp 4, sourcemap support is built into Gulp. Switch over to this new method (https://gulpjs.com/docs/en/api/src#sourcemaps).

This will help us remove a dependency on gulp-sourcemaps, which is pinning a version of postcss with a known vulnerability.

Copied from https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1448

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -5,7 +5,6 @@ const sass = require('gulp-sass')
 const filelog = require('gulp-filelog')
 const include = require('gulp-include')
 const path = require('path')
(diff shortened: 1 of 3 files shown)


gidsg (approved):
LGTM ℹ️ 👍

